PR #14, opened 2016-06-23
Got invliad uri on android, and found this can work on android

wasabia:
http://stackoverflow.com/questions/19834842/android-gallery-on-kitkat-returns-different-uri-for-intent-action-get-content

http://stackoverflow.com/questions/26355547/android-application-crash-when-picking-photo-on-new-devices


ivpusic:
will close this one in favor of https://github.com/ivpusic/react-native-image-crop-picker/pull/17. Tnx for PR

PR #17, opened 2016-07-03
Use better function to get real path

superandrew213:
Android gallery on KitKat returns different uri

```
content://media/external/images/media/3951

content://com.android.providers.media.documents/document/image:3951
```


nitsujw:
+1 - I've tested this change and it fixes crashing the app and returns the proper path

ivpusic:
tnx guys. will try it today

PR #35, opened 2016-08-05
Add base64 data in result 

magrinj:
Add the base64 image data, in the response returned

### DIFF
--- a/android/src/main/java/com/reactnative/picker/PickerModule.java
+++ b/android/src/main/java/com/reactnative/picker/PickerModule.java
@@ -53,6 +53,7 @@ public class PickerModule extends ReactContextBaseJavaModule implements Activity
 
     private boolean cropping = false;
     private boolean multiple = false;
+    private boolean includeBase64 = false;
     private int width = 100;
     private int height = 100;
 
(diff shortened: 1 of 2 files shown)


ivpusic:
hmm...is there really good use-case for this? Usually when you pick image you want to show it to the user, what you can right now, or you want to upload it to the server, what you can right now. What is your use case for introducing base64?

ivpusic:
when you get filepath you can (optionally) read base64 with https://github.com/johanneslumpe/react-native-fs.

magrinj:
@ivpusic With react-native you can set a picture with base64 datas, this case is a special case when some user want to send datas to server just after the image is selected. I think it's an enhancement for your library and lot of people can be in this case.

Base64 datas can also be used with `AsynStorage` to save a picture locally.

I try to use `react-native-fs` but that's doesn't work on my project, and I don't really want to use a big library just to get datas that are already get on the native side.

ivpusic:
my concern are memory issues on android devices. I already had issue with decoding bitmap (https://github.com/ivpusic/react-native-image-crop-picker/blob/master/android/src/main/java/com/reactnative/picker/PickerModule.java#L110). I think by default we should not send base64, because for most use cases you don't need it, and you will waste phone memory.

I suggest that if we want to support this feature, to make it optional. So when selecting image if user wants to get base64, it should provide flag `includeBase64: true` to `openPicker` method.

What do you think? If you agree, can you update your PR to make this feature optional?

magrinj:
I update my PR with the includeBase64 option as you suggest, I use `NSMutableDictionnary` for the `iOS` side, I don't really use objective-C often, so if you think there is a better solution just let me know.

ivpusic:
thanks. I did some small changes and now it is merged. Will release new version

ivpusic:
https://github.com/ivpusic/react-native-image-crop-picker/releases/tag/v0.5.5

magrinj:
Thanks for your merge and your change!

PR #36, opened 2016-08-05
Magrinj base64 data

ivpusic:


ivpusic:
original PR https://github.com/ivpusic/react-native-image-crop-picker/pull/35

PR #41, opened 2016-08-10
added camera support for react-native-image-crop-picker

sunnyyants:
I added the camera support of this package, which is base on the implementation of react-native-image-picker. I added another function call openCamera which basically allows user to launch the camera once the permission is granted. In the mean time the user also can continue to crop or rotate the image after finished taking photo from camera. I used the same options as openPicker.

### DIFF
--- a/android/src/main/java/com/reactnative/picker/PickerModule.java
+++ b/android/src/main/java/com/reactnative/picker/PickerModule.java
@@ -12,6 +12,8 @@ import android.os.Build;
 import android.provider.DocumentsContract;
 import android.provider.MediaStore;
 import android.util.Base64;
+import android.Manifest;
+import android.os.Environment;
 
 import com.facebook.react.bridge.ActivityEventListener;
 import com.facebook.react.bridge.Promise;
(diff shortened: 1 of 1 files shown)


ivpusic:
Hey, PR looks good. However I would like to keep this package as iOS/Android compatible, so in this case we are missing iOS part. Do you have a time to take a look into iOS part as well?

chiraggshah:
It would be a great help if the iOS part is implemented. Camera support would be a major thing to have.

sunnyyants:
Ok, let me see whether I could contribute to the iOS part as well, it might take some time since I am not quite familiar with iOS part, but I could have a try~

ivpusic:
@sunnyyants I will do it. will merge this with iOS part

ivpusic:
Merged in https://github.com/ivpusic/react-native-image-crop-picker/pull/44. Thanks a lot for contribution. Camera feature available in version 0.6.0

chiraggshah:
@ivpusic : That was fast..! Thanks a lot.

PR #44, opened 2016-08-11
[FEATURE] Camera support

ivpusic:


BilalBudhani:
@ivpusic @sunnyyants This is amazing! Thanks for all the hard work you guys put in behind all this. The least we can do is show our appreciation. 👍

sunnyyants:
@ivpusic, I think I forgot to mention that it seems we have to add some extra config on AndroidManifest.xml, which is adding some camera related permission on it to make sure the camera is accessible.

`<uses-permission android:name="android.permission.CAMERA" />
    <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"/>
    <uses-feature android:name="android.hardware.camera" android:required="false"/>
    <uses-feature android:name="android.hardware.camera.autofocus" android:required="false"/>`

ivpusic:
probably it is better not to make it required, because camera picker is optional...if you are going to use it in your project you should add it to your project manifest file

android/src/main/java/com/reactnative/picker/PickerModule.java
```diff
@@ -40,13 +46,18 @@
...
 
     private static final int IMAGE_PICKER_REQUEST = 1;
+    private static final int CAMERA_PICKER_REQUEST = 2;
     private static final String E_ACTIVITY_DOES_NOT_EXIST = "E_ACTIVITY_DOES_NOT_EXIST";
 
     private static final String E_PICKER_CANCELLED_KEY = "picker_cancel";
     private static final String E_PICKER_CANCELLED_MSG = "User cancelled image selection";
 
     private static final String E_FAILED_TO_SHOW_PICKER = "E_FAILED_TO_SHOW_PICKER";
+    private static final String E_FAILED_TO_OPEN_CAMERA = "E_FALIED_TO_OPEN_CAMERA";
```
Nicolas-Menettrier: Typing error : "E_FALIED_TO_OPEN_CAMERA" in place of "E_FAILED_TO_OPEN_CAMERA"

ivpusic: thanks. will correct this

PR #67, opened 2016-09-09
added resizeMultiple option to reduce size of base64 response

parkerdan:
resize base64 on multi-select, changed Android default height/width to match iOS.


ivpusic:
@parkerdan thank you for your contribution. However for now I would like to keep cropping feature only if you take single photo. 

This PR could make sense for your specific case, but usually you want to see how result image will look like before cutting it, and in this case we would just cut it.

PR #68, opened 2016-09-10
Add smart album option iOS only

superandrew213:
Users can specify which smart albums to show and/or reorder.


ivpusic:
@superandrew213 thank you for PR. +1 for feature, but I have question. Why `smartAlbums` and not just simple `albums`? Can you rename this, and update variables/comments/readme to match simpler name?

Also would be cool if you could try to resolve merge conflicts which came after commit for video support https://github.com/ivpusic/react-native-image-crop-picker/commit/457c0efbd5c9e9c6901fd74cf27336a8d601da0f.

superandrew213:
Ok will fix conflicts soon.

I used smart albums because that's what their are called on iOS and to make a distinction between smart albums and albums the user creates. This option only controls the smart albums. The user created albums will always show.

Let me know what you think.

ivpusic:
+1 then for `smartAlbums`. Just please update readme with default behaviour for `smartAlbums` (what will happen if option is not specified)

superandrew213:
I have added the default behaviour already in the "Request Object" section in the readme.

array ['UserLibrary', 'PhotoStream', 'Panoramas', 'Videos', 'Bursts']

Where else do you want me to mention it?

ivpusic:
just format it like this to be consistent with rest of the request object table -> `array (default ['UserLibrary', 'PhotoStream', 'Panoramas', 'Videos', 'Bursts'])`. This is also not fully correct, because default is to skip setting this option, but it is ok because these are all possible options.

superandrew213:
Sorry my bad, fixed it

ivpusic:
tnx a lot!

PR #87, opened 2016-09-27
Icloud

magrinj:
Ok, I find why icloud image are not working! 
To solve this issue we need to specify in `PHImageRequestOptions` the option `networkAccessAllowed`, it's what I did right here.
But I still have a problem, the activity indicator never appear after I click on a picture :/ 
I try to use the function created and put it in the main thread but still not working, any ideas ?
Instead of a spinner we can also print a ProgressBar with this handler:

```
                options.progressHandler = ^(double progress, NSError *error, BOOL *stop, NSDictionary *info) {
                    NSLog(@"%f", progress);
                };
```

### DIFF
--- a/ios/ImageCropPicker.m
+++ b/ios/ImageCropPicker.m
@@ -430,23 +430,25 @@ RCT_EXPORT_METHOD(openPicker:(NSDictionary *)options
                 }];
             }];
         } else {
-            [self showActivityIndicator:^(UIActivityIndicatorView *indicatorView, UIView *overlayView) {
-                options.progressHandler = ^(double progress, NSError *error, BOOL *stop, NSDictionary *info) {
-                    NSLog(@"%f", progress);
-                    if (stop == YES) {
-                        [indicatorView stopAnimating];
-                        [overlayView removeFromSuperview];
-                    }
-                };
-                [manager
-                 requestImageDataForAsset:phAsset
-                 options:options
-                 resultHandler:^(NSData *imageData, NSString *dataUTI,
-                                 UIImageOrientation orientation,
-                                 NSDictionary *info) {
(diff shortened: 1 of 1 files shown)


ivpusic:
as I don't have iCloud images to test it is harder to help you. What happens if you put `options.synchronous = NO;`?

magrinj:
I don't try to set `options.synchronous = NO;`, as the doc suggest I don't think this option is going to change something for icloud pictures: https://developer.apple.com/reference/photos/phimagerequestoptions
With the option I add, all the pictures work, even the icloud pictures, the problem is just that if I click on an icloud picture there is a delay before the view pop and push the cropping view, the delay is due to the download of the image. My problem is that I can't print the spinner on the screen when the image is downloaded, I try to use your function with the handler, but the spinner is never shown. Did you know why ?

ivpusic:
correct, and this is why I suggested to try `options.synchronous = NO;`, because maybe main thread with `options.synchronous = YES;` is blocked and it cannot show spinner

magrinj:
Oh yes you're right, I'm going to try this!

magrinj:
Ok sound good now, the spinner appear! :)

ivpusic:
tnx. I also pushed fix for iOS 10 (for example project), so I will close this PR in favor of this commit https://github.com/ivpusic/react-native-image-crop-picker/commit/d6e2f7d2d13a68ddf11a801fd13870eecbc8a124.

magrinj:
Ok perfect! Thanks for the merge! 👍

superandrew213:
Thanks for getting this fixed @magrinj 👍

magrinj:
You're welcome! :)

PR #109, opened 2016-10-13
Changed permission check to use react native request permission and a…

nickythorne:
The existing permission check wasn't handling the asynchronous request properly and therefore resulted in a permission denied error even when the permissions were granted. I have changed this to use React Native's version of the request permission mechanism allowing the module to register a permission listener and continue with the camera invocation after this point.

Additionally, the module now also requests the `WRITE_EXTERNAL_STORAGE`  permission, as this is required by uCrop at runtime.


ivpusic:
Thank you for this PR. I refactored it a little bit and added permission check when selecting image without camera, and when trying to clear tmp images. https://github.com/ivpusic/react-native-image-crop-picker/commit/d6c023fbed8ffb5bdcee74f363311049026ce8f9

Issue #111, opened 2016-10-14
Provide more unique ActivityEvenListener requestCodes


kylejones:
### Version
Tell us which versions you are using: 

- react-native-image-crop-picker v0.10.0
- react-native v0.35.0

### Platform
Tell us to which platform this issue is related

- Android

### Steps to reproduce

I'm running react-native-image-picker and react-native-braintree-xplat. Both have ActivityEventListeners with a requestCode of 1. The collision in requestCodes causes app crashes when one module tries to handle the request from the other. I would suggest using a large random integer to reduce the likelihood of collisions with other modules.

https://github.com/ivpusic/react-native-image-crop-picker/blob/master/android/src/main/java/com/reactnative/ivpusic/imagepicker/PickerModule.java#L48

Stack trace
```sh
java.lang.RuntimeException: Failure delivering result ResultInfo{who=null, request=1, result=-1, data=Intent { dat=content://com.android.providers.downloads.documents/document/60 flg=0x1 }} to activity {com.seejanego/com.seejanego.MainActivity}: java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.String com.braintreepayments.api.models.PaymentMethodNonce.getNonce()' on a null object reference
    at android.app.ActivityThread.deliverResults(ActivityThread.java:3699)
    at android.app.ActivityThread.handleSendResult(ActivityThread.java:3742)
    at android.app.ActivityThread.-wrap16(ActivityThread.java)
    at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1393)
    at android.os.Handler.dispatchMessage(Handler.java:102)
    at android.os.Looper.loop(Looper.java:148)
    at android.app.ActivityThread.main(ActivityThread.java:5417)
    at java.lang.reflect.Method.invoke(Native Method)
    at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:726)
    at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:616)
 Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.String com.braintreepayments.api.models.PaymentMethodNonce.getNonce()' on a null object reference
    at com.pw.droplet.braintree.Braintree.onActivityResult(Braintree.java:116)
    at com.facebook.react.bridge.ReactContext.onActivityResult(ReactContext.java:226)
[...]

PR #112, opened 2016-10-15
Setting ActivityEventListener requestCodes to large ints to prevent c…

kylejones:
- Setting ActivityEventListener requestCodes to large ints to prevent collisions with other modules


kylejones:
References #111

ivpusic:
tnx for this

PR #128, opened 2016-11-05
add if (mPickerPromise == null) to croppingResult 

ksti:
should check if (mPickerPromise == null) first in croppingResult  just like other result calling:imagePickerResult and cameraPickerResult. Because when app's mainActivity resumed and called onActivityResult can you get the mPickerPromise.


ivpusic:
Thanks

PR #133, opened 2016-11-10
Supporting compressing return based64 image data

Jaafar-abusair:
Now by adding isCompressedBase64 to ImagePicker2.openPicker options, you will always get compressed based64 data around 150-200k only.

Note that the includeBase64 should be true and the default value for isCompressedBase64 is false,
If someone need to specify the compressed data size, I will support that.

 ImagePicker2.openPicker({
            multiple: true,
            includeBase64: true,
            isCompressedBase64: true,
.....
})


ivpusic:
tnx for pull request. Android side looks good, but as we have base64 feature on iOS side as well, would be great if you can do it there also?

Jaafar-abusair:
I have no idea how to do it for IOS :p no experience with that

PR #140, opened 2016-11-14
Allow choice of default camera (front vs. back) on iOS

theseansy:
For easily setting an avatar picture, it is nice to be able to choose the front facing camera as the default.

This is trivial on iOS, on Android it's [a little more complicated](http://stackoverflow.com/questions/2779002/how-to-open-front-camera-on-android-platform) so implemented just for iOS for now.

### DIFF
--- a/ios/ImageCropPicker.m
+++ b/ios/ImageCropPicker.m
@@ -42,7 +42,8 @@ RCT_EXPORT_MODULE();
                                 @"compressVideo": @YES,
                                 @"maxFiles": @5,
                                 @"width": @200,
-                                @"height": @200
+                                @"height": @200,
+                                @"useFrontCamera": @NO
                                 };
     }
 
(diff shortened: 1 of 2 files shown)


ios/ImageCropPicker.m
```diff
@@ -106,7 +107,9 @@ - (UIViewController*) getRootVC {
         picker.delegate = self;
         picker.allowsEditing = NO;
         picker.sourceType = UIImagePickerControllerSourceTypeCamera;
-        picker.delegate = self;
```
theseansy: Delegate already was set above ^

ivpusic:
tnx a lot. do you have a time to do the same for android?

theseansy:
I can give it a go tonight.  There may be some weird edge cases for cameras with no front-facing cameras.

If this is the case, i'll just ignore the `useFrontCamera` flag.

theseansy:
@ivpusic - Turned out being much less trivial than anticipated.  There are a lot of changes around this between different versions of android and how to choose the camera. For example, [this method](http://stackoverflow.com/questions/2779002/how-to-open-front-camera-on-android-platform) is now outdated as it uses older Android-deprecated APIs.

Finally, some android devices remember the camera used (will default to front-facing if last picture was taken with a front-facing camera). This is a slight deviation from iOS where that setting is never persisted.  I'd recommend just merging this, and if there is demand for android, put it in later.

ivpusic:
thanks a lot

PR #146, opened 2016-11-17
Fixed bug introduced where user can't cancel out of cropping image

theseansy:


theseansy:
resubmitting, branched incorrectly.

PR #147, opened 2016-11-17
Fixed bug where user couldn't cancel out of cropping (ios)

theseansy:
I introduced this bug, but here's the fix


ivpusic:
Thanks man

PR #148, opened 2016-11-18
Allow user to specify brand color for cropper

theseansy:
- Default colors for the cropper are less aggressive (not UCrop's orange)
- If cropperTintColor is passed in, this overrides status bar, toolbard, and widget UX items.

Default colors if no `cropperTintColor` passed in:
screenshot_1479431752


ivpusic:
thanks man

PR #150, opened 2016-11-21
Activity must be cast to PermissionAwareActivity

chaitanya0bhagvan:
Activity must be cast to PermissionAwareActivity instead of ReactActivity. This was a problem when using with react-native-navigation since it does not extend ReactActivity


ivpusic:
Thanks. Merged and released in https://github.com/ivpusic/react-native-image-crop-picker/releases/tag/v0.10.6

PR #161, opened 2016-12-08
onChooseButtonTouch disenable the button to avoid callbacking twice in RN. 

ksti:
which could cause a red screen issue: only one callback maybe registered to a function in a native module

img_0499


ivpusic:
hmm, I am not in charge or external libraries, so it is better to submit PR to original project page if you have ideas for improvement.

ksti:
I will submit it to the original project, but it is better you also merge this because we met an issue on iOS using your library when openCamera() and callback. It would crash if click "Choose" twice! And it is more often click twice because of no response before callback.

ivpusic:
to prevent this you can "disable" button until promise resolves or rejects.

PR #167, opened 2016-12-16
add support to uCrop setCircleDimmedLayer()

esganzerla:
set the setCircleDimmedLayer to "cropperCircleOverlay" option on react-native.


ivpusic:
Thanks for PR. could you do the same for ios?

esganzerla:
Cool. I just included the same property for iOS.

ivpusic:
Awesome. Thanks!

WillyRamirez:
Hi guys, thanks for this awesome feature. Question: does this provide only a circular overlay during cropping, or will the image also be cropped in a circle? So far with cropperCircleOverlay set to true, im still getting square images. Thanks in advance!

Issue #184, opened 2017-01-03
enable cropping on png pic with transparent background returns pic with white background color.


DoTalkLily:
ImagePicker.openPicker({
            maxWidth: 720,
            maxHeight: 720,
            width: 240,
            height: 108,
            cropping: true
        }).then(image => {});
After cropping my png image which background color is transparent, I got a image width white background.  How to fix it?

ivpusic:
all images are converted to jpg. If you are interested to make this dependent on selected image mime type, I would accept PR.

PR #197, opened 2017-01-13
fix issue #184 add image mime type detection

DoTalkLily:
https://github.com/ivpusic/react-native-image-crop-picker/issues/184

### DIFF
--- a/ios/Compression.m
+++ b/ios/Compression.m
@@ -62,13 +62,18 @@
                    withOptions:(NSDictionary*)options {
     ImageResult *result = [self compressImageDimensions:image withOptions:options];
     
-    NSNumber *compressQuality = [options valueForKey:@"compressImageQuality"];
-    if (compressQuality == nil) {
-        compressQuality = [NSNumber numberWithFloat:1];
+    if([@"image/png" isEqualToString:[options objectForKey:@"mimeType"]]){
+        result.data = UIImagePNGRepresentation(image);
+        result.mime = @"image/png";
+    }else{
+        NSNumber *compressQuality = [options valueForKey:@"compressImageQuality"];
+        if (compressQuality == nil) {
+            compressQuality = [NSNumber numberWithFloat:1];
+        }
+        
+        result.data = UIImageJPEGRepresentation(result.image, [compressQuality floatValue]);
+        result.mime = @"image/jpeg";
     }
-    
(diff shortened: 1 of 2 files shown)


ivpusic:
thanks for PR. Feature is great, but I suggest to change one thing. We should not put mime type into options. Instead for every image we should extract mime type and convert it to png and jpg, depending on mime type. So in this case, we would do this in Compression.m. All code regarding storing mime type to options should be removed.

DoTalkLily:
You're right. But one more thing, the function should be called with NSData, which I could got in :
[manager
                 requestImageDataForAsset:phAsset
                 options:options
                 resultHandler:^(NSData *imageData, NSString *dataUTI,
                                 UIImageOrientation orientation,
                                 NSDictionary *info) {
                     dispatch_async(dispatch_get_main_queue(), ^{
                         [indicatorView stopAnimating];
                         [overlayView removeFromSuperview];
                         [self processSingleImagePick:[UIImage imageWithData:imageData] withViewController:imagePickerController];
                     });
                 }];

Or I have to call UIImageJPEGRepresentation to get NSData to determine the mime type.
Any suggestions ?

Oh, I miss the function:
persistFile, where all tmp pic is stored with 'jpg' suffix.

ivpusic:
we could try to extend compressImage method with `withMime` parameter, and then before passing try to figure out mime type. I don't know if this is possible with PHAsset, some investigation is needed here. Could you do that? Thanks

DoTalkLily:
hi, in this callback function
 - (void)imageCropViewController:(RSKImageCropViewController *)controller
                   didCropImage:(UIImage *)croppedImage
                  usingCropRect:(CGRect)cropRect 
We got the UIImage ,I didn't find a way to get mime type or file path directly from UIImage. One workaround is to get NSData with UIImagePNGRepresentation then detect the type. (If use UIImageJPEGRepresentation the - (NSString *)mimeTypeForData:(NSData *)data  will be jpeg.)

image

DoTalkLily:
An other way to make code neater is to use associated objects mechanism . We could store the mime type in UIImage as a feature of the image, it seems to be more appropriate. What you think?

ivpusic:
I can see things will be starting more and more complicated to achieve the simple goal. Currently I don't see very clean way to implement detecting mime type detection for all our options. For now, I will close this PR.

Thanks a lot for your time and effort.

PR #199, opened 2017-01-17
support ios 8

luhui:


ivpusic:
thanks

ivpusic:
actually, I think it makes more sense to do this check from your application code. You will anyway have to specify `compressVideoPreset` from code. If your device does not support 4k, put some other, or do condition in your code.

ivpusic:
update: PR reverted

luhui:
Hi, the crash is in the `init` method of `Compression`, `AVAssetExportPreset3840x2160` is available 9.0 above.   
So check the system version in the `init` method is necessary

Here is definition in AVAssetExportSession.h
image

ivpusic:
ahaa, then sorry, my bad, will merge it back

PR #220, opened 2017-01-26
Fix for unknown imports

danjersoft:
After trying the instructions multiple times, I wasn't able to get anywhere until I tried the following:

http://stackoverflow.com/a/41664041/752815

This helped as well:

https://medium.com/@thisismissem/how-to-upgrade-react-native-modules-in-a-backwards-compatible-manner-a5b5c48d590c#.eyvdfp97f


ivpusic:
thanks for PR. You are getting errors because you are using RN version < 0.40. Upgrade to 0.40 and you won't get these issues. Your changes will make it backwards compatible. Thanks!

Issue #235, opened 2017-02-14
Failure delivering result ResultInfo : Illegal callback invocation from native module. This callback type only permits a single invocation from native code


the-simian:
### Version
- react-native-image-crop-picker v0.12.2
- react-native v0.41.2
- react v15.4.2

### Platform
- Android

### Trace
```sh
java.lang.RuntimeException Failure delivering result ResultInfo{who=null, request=69, result=0, data=null} to activity {com.seejanego/com.seejanego.MainActivity}: java.lang.RuntimeException: Illegal callback invocation from native module. This callback type only permits a single invocation from native code. 
    ActivityThread.java:4925 android.app.ActivityThread.deliverResults
    ActivityThread.java:4968 android.app.ActivityThread.handleSendResult
    ActivityThread.java:222 android.app.ActivityThread.access$1600
    ActivityThread.java:1849 android.app.ActivityThread$H.handleMessage
    Handler.java:102 android.os.Handler.dispatchMessage
    Looper.java:158 android.os.Looper.loop
    ActivityThread.java:7229 android.app.ActivityThread.main
    Method.java:-2 java.lang.reflect.Method.invoke
    ZygoteInit.java:1230 com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run
    ZygoteInit.java:1120 com.android.internal.os.ZygoteInit.main


Caused: java.lang.RuntimeException Illegal callback invocation from native module. This callback type only permits a single invocation from native code. 
    CallbackImpl.java:32 com.facebook.react.bridge.CallbackImpl.invoke
    PromiseImpl.java:70 com.facebook.react.bridge.PromiseImpl.reject
    PromiseImpl.java:38 com.facebook.react.bridge.PromiseImpl.reject
    ResultCollector.java:65 com.reactnative.ivpusic.imagepicker.ResultCollector.notifyProblem
    PickerModule.java:619 com.reactnative.ivpusic.imagepicker.PickerModule.croppingResult
    PickerModule.java:630 com.reactnative.ivpusic.imagepicker.PickerModule.onActivityResult
    ReactContext.java:261 com.facebook.react.bridge.ReactContext.onActivityResult
    XReactInstanceManagerImpl.java:627 com.facebook.react.XReactInstanceManagerImpl.onActivityResult
    ReactActivityDelegate.java:149 com.facebook.react.ReactActivityDelegate.onActivityResult
[...]

PR #237, opened 2017-02-16
Add synchronized to notifyProblem

the-simian:


the-simian:
fixes https://github.com/ivpusic/react-native-image-crop-picker/issues/235

Issue #177, opened 2016-12-29
Image picker not wokring


navneetccna:
### Version
Tell us which versions you are using: 
- "react-native-image-crop-picker": "^0.10.9"
- "react-native": "^0.23.0",

### Platform
Tell us to which platform this issue is related
- Android

### Expected behaviour
Open image picker

import ImagePicker from 'react-native-image-crop-picker';
ImagePicker.openPicker({
		  width: 300,
		  height: 400,
		  cropping: true
		}).then(image => {
		  console.log(image);
		});

### Actual behaviour
Error : undefined is not an object (evaluating '_reactNativeImageCropPicker2.default.openPicker')

I also tested in the android studio and getting this error 

Error:(31, 39) error: cannot find symbol class PermissionAwareActivity
Error:(32, 39) error: cannot find symbol class PermissionListener
Error:(47, 1) error: PickerModule is not abstract and does not override abstract method onActivityResult(int,int,Intent) in ActivityEventListener
Error:(208, 143) error: cannot find symbol class PermissionListener
Error:(208, 15) error: cannot find symbol class PermissionAwareActivity
Error:(585, 5) error: method does not override or implement a method from a supertype
Error:(596, 5) error: method does not override or implement a method from a supertype
Error:Execution failed for task ':react-native-image-crop-picker:compileReleaseJavaWithJavac'.
> Compilation failed; see the compiler error output for details.

ivpusic:
it seems that you did not install it correctly. Please follow all steps from here https://github.com/ivpusic/react-native-image-crop-picker#install

navneetccna:
I followed all the steps even i tried again and again multiple times but same issue

navneetccna:
Last installation log is 

npm WARN react-native-image-crop-picker@0.10.9 requires a peer of react-native@>=0.33.0 but none was installed.
npm WARN react-native-pulse-loader@1.0.2 requires a peer of react-native@^0.27 but none was installed.
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: bufferutil@1.2.1 (node_modules\bufferutil):
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: bufferutil@1.2.1 install: `node-gyp rebuild`
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: Exit status 1
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: utf-8-validate@1.2.2 (node_modules\utf-8-validate):
[...]

PR #245, opened 2017-02-23
#177 Update build.gradle to reference node_modules for react-native dependency

TigerC10:


ivpusic:
hm, could you explain why this change?

TigerC10:
It was discussed in #177 but the gist of it is that in the build.gradle file you specify a dependency on the latest version of `com.facebook.react:react-native`.  This dependency is satisfied through maven.

If someone has downloaded react-native and built it themselves before including this project, their local .m2 cache will have an appropriate version of react-native.

However, if you haven't done that separate step before it'll attempt to download react-native from maven central.  Facebook stopped pushing updates to maven central, the latest version in that maven repo is `0.20.1` which does not have the requisite library `com.facebook.react.modules.core.PermissionListener`.

Now, the consuming projects out there that leverage react-native-image-crop-picker could possibly configure the gradle build to make sure that react-native compiles first and feed that into react-native-image-crop-picker's build phase...  But it would add some extra configuration to the installation steps and doesn't really fix Android Studio's tendency to move react-native-image-crop-picker build before the react-native build when "upgrading" gradle files.

A more resilient solution is to add the local copy of react-native from the `node_modules` directory as a maven repository.  This also guarantees that you won't ever use a mis-matched version of react-native to build react-native-image-crop-picker, it'll always use the version installed in the project by `npm`.

TigerC10:
@ivpusic - can you think of an alternate fix?

TigerC10:
@ivpusic any updates?

sibelius:
I don't think this is needed for this package?

TigerC10:
It's pretty common for react-native modules that depend on a specific version of react-native in android redefine the repositories such that the RN android dependency checks inside the node_modules directory.  Check https://github.com/facebook/react-native/issues/12309 for the details.

Here are some other projects doing it:
https://github.com/airbnb/react-native-maps/blob/master/build.gradle
https://github.com/devfd/react-native-google-signin/blob/master/android/build.gradle
https://github.com/Nodonisko/react-native-twilio/blob/develop/android/build.gradle


Though, another method of achieving this might be to use [react.gradle](https://github.com/facebook/react-native/blob/master/react.gradle) instead:
```
apply from: "../../node_modules/react-native/react.gradle"
```

Palisand:
@TigerC10 I'm having the same issue with [react-native-image-picker](https://github.com/react-community/react-native-image-picker/blob/develop/android/build.gradle) after building RN from source. I haven't been able to get it working; care to lend a hand?

Palisand:
@sibelius You're probably right about this PR not being necessary. After experiencing the same android build failure issues with react-native-image-picker and react-native-maps, I found [this solution](https://github.com/airbnb/react-native-maps/issues/1309#issuecomment-323689000).

If you're building RN from source and come across this issue, updating your app's _`android/build.gradle`_ with the following should do the trick:
```
allprojects {
    // ...
    configurations.all {
        resolutionStrategy {
            dependencySubstitution {
                substitute module('com.facebook.react:react-native') with project(':ReactAndroid')
            }
        }
    }
}
```
Credit: @madox2

ivpusic:
closing it at this is not necessary

subooom:
`project :ReactAndroid not found.`
After copy pasting this:

     configurations.all {
        resolutionStrategy {
            dependencySubstitution {
                substitute module('com.facebook.react:react-native') with project(':ReactAndroid')
            }
        }
    }

devotebest:
I got same project :ReactAndroid not found.

PR #247, opened 2017-02-24
iOS: Wait for crop VC dismiss before resolve

benvium:
Allows ImagePicker.openCamera().then() handler to perform Alert.alert.

Test case:

```objc
ImagePicker.openCamera({
    compressImageMaxWidth: 1024,
    compressImageMaxHeight: 1024,
    compressImageQuality: 0.7,
  })
    .then((image) => {
      Alert.alert('Picture taken')
    });
```

In existing codebase
--
- The Alert is not displayed.  Repeat with a Chrome breakpoint on the Alert line - wait a few seconds, then continue = Alert displayed. Demonstrates this is a timing issue

With this PR
----

- the alert is displayed every time, regardless of whether there's a breakpoint or not.

### DIFF
--- a/ios/ImageCropPicker.m
+++ b/ios/ImageCropPicker.m
@@ -522,14 +522,16 @@ RCT_EXPORT_METHOD(openCropper:(NSDictionary *)options
             return;
         }
 
-        self.resolve([self createAttachmentResponse:filePath
-                                          withWidth:imageResult.width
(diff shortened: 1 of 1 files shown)


benvium:
Closing as I need applied to master rather than release 0.12.4. Will re-send in new PR

PR #248, opened 2017-02-24
iOS: Wait for viewController to dismiss before resolving or rejecting

benvium:
Allows ImagePicker.openCamera().then() handler to perform Alert.alert.

Test case:

```objc
import {Alert} from 'react-native';

ImagePicker.openCamera({
    compressImageMaxWidth: 1024,
    compressImageMaxHeight: 1024,
    compressImageQuality: 0.7,
  })
    .then((image) => {
      Alert.alert('Picture taken')
    });
```

In existing codebase
--
- The Alert is not displayed.  Repeat with a Chrome breakpoint on the Alert line - wait a few seconds, then continue = Alert displayed. Demonstrates this is a timing issue

With this PR
----

- the alert is displayed every time, regardless of whether there's a breakpoint or not.

The same logic has been applied to each .reject() too.

Tested via Xcode 8.2.1, React Native 0.41.2 on iPhone 6 running iOS 10.2

### DIFF
--- a/ios/ImageCropPicker.m
+++ b/ios/ImageCropPicker.m
@@ -604,7 +604,7 @@ RCT_EXPORT_METHOD(openCropper:(NSDictionary *)options
     }];
 }
 
-- (void) dismissCropper:(RSKImageCropViewController*) controller completion:(^void) completion{
+- (void) dismissCropper:(RSKImageCropViewController*) controller completion:(void (^)())completion {
     //We've presented the cropper on top of the image picker as to not have a double modal animation.
     //Thus, we need to dismiss the image picker view controller to dismiss the whole stack.
     if (!self.cropOnly) {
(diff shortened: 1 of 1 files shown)


ivpusic:
thanks for PR. Before merging this, we should ensure consistent behavior every time image is returned to the caller. Currently, it will have the possibility to see alert only if he is selecting a single image. Could you look for other places where we are returning the result to the caller, and add completion block (maybe it makes sense just to extract this completion handler into the separate function, and just use it everywhere). 

Basically, you can just ctrl+f for `self.resolve` to find where are we returning something to the caller.

benvium:
@ivpusic Yes, that's a good idea. I'll check and update.

ivpusic:
awesome, thanks man

Issue #259, opened 2017-03-06
[Android] Can't import from SD card


superandrew213:
### Version
Tell us which versions you are using: 

- react-native-image-crop-picker v0.11.2
- react-native v0.39.2

### Platform
Tell us to which platform this issue is related

- Android

### Expected behaviour

Be able to import from SD card.

### Actual behaviour

You can browse SD card but no media is returned.

ivpusic:
I would like to help here, but on all devices I have, this works :/. Would be very cool if you can dig deeper into this.

superandrew213:
@ivpusic are you sure it's working? It's not even working on the sim for me.

ivpusic:
oh, would be great if I could reproduce this on simulator. Which simulator are you using? What version?

superandrew213:
<img width="494" alt="screen shot 2017-03-06 at 7 15 46 pm" src="https://cloud.githubusercontent.com/assets/13059204/23603548/67b8676e-02a1-11e7-9030-eefcc15e5248.png">

ivpusic:
could you try new version https://github.com/ivpusic/react-native-image-crop-picker/releases/tag/v0.12.7.

superandrew213:
Didn't work for me. Is it working for you?

ivpusic:
hm, yes, works for me.

superandrew213:
You testing with same sim settings?

ivpusic:
same same

superandrew213:
Type [here](https://github.com/ivpusic/react-native-image-crop-picker/commit/c483a3f2f095c42522c20c39e41ac1f486ee675d#diff-bb02fac60d336bc591a1f78314ce0a2eL25) equals 0A15-2C09

ivpusic:
for me as well, but split[1] was correct path for me, so I returned `return Environment.getExternalStorageDirectory() + "/" + split[1];`, and it was valid file path which points to real file, and file was shown on react-native app in the end. I was googling a little bit about that, but didn't find anything useful so far (it was 5min googling).

For you `return Environment.getExternalStorageDirectory() + "/" + split[1];` created invalid path or?

superandrew213:
It does give me a path but I also get this ```image-crop-picker: Promise rejected. Cannot resolve asset path.```

ivpusic:
hm, that means path was null or empty, but as you said you get some path, and this error should not happen, even if path is not valid => it should not be null or empty.

[...]

PR #262, opened 2017-03-07
Fix resolving external storage paths

superandrew213:
https://github.com/ivpusic/react-native-image-crop-picker/issues/259


ivpusic:
thank a lot! Released here https://github.com/ivpusic/react-native-image-crop-picker/releases/tag/v0.12.8

PR #290, opened 2017-04-04
Add typescript definition file

wcandillon:
Thanks for this great module. This is the definition file I wrote in order to use it with ts.


ivpusic:
tnx!

Issue #283, opened 2017-03-30
Android 4.4.4 can't open camera


arthasAndDk:
### Version
Tell us which versions you are using: 
- react-native-image-crop-picker v0.10.5
- react-native v0.39.2

### Platform
Tell us to which platform this issue is related
- Android 4.4.4

### Detail
PickerModule.java

private File createImageFile() throws IOException {
        String imageFileName = "image-" + UUID.randomUUID().toString();
        File path = Environment.getExternalStoragePublicDirectory(
                Environment.DIRECTORY_PICTURES);
        File image = File.createTempFile(imageFileName, ".jpg", path);

        // Save a file: path for use with ACTION_VIEW intents
        mCurrentPhotoPath = "file:" + image.getAbsolutePath();

        return image;
}

run on Android 4.4.4

if Environment.DIRECTORY_PICTURES is not exist
File.createTempFile unable to create folder automatically

so you should create Environment.DIRECTORY_PICTURES by yourself

andrevargas:
Hello, I'm having the same issue. The JS method `ImagePicker.openCamera` doesn't produce any effect on **Android 4.4**, and after debugging with AndroidStudio I've reached the same problem reported above. Some details about my case:

#### Version
* react-native-image-crop-picker v0.12.9
* react-native 0.40.0

#### Platform
* Android 4.4 (KitKat), API level 19

#### Details
As said by @arthasAndDk, if the directory pointed by `Environment.DIRECTORY_PICTURES` doesn't exist in the file system, the method `File.createTempFile()` throws an `IOException: open failed: ENOENT (No such file or directory)`. So, for testing purposes, my code looks like this now:

```java
private File createImageFile() throws IOException {

    String imageFileName = "image-" + UUID.randomUUID().toString();
    File path = Environment.getExternalStoragePublicDirectory(
        Environment.DIRECTORY_PICTURES);
    
    // Create the directory if it doesn't exist
    if (!path.exists() && !path.isDirectory()) {
        path.mkdirs();
    }

    File image = File.createTempFile(imageFileName, ".jpg", path);

    mCurrentPhotoPath = "file:" + image.getAbsolutePath();

    return image;

}
```
...And everything works fine! I'm opening a PR with these fixes.

By the way, thanks for this awesome library, and thank you @arthasAndDk for opening and explaining this issue! :octocat:

ivpusic:
fixed in https://github.com/ivpusic/react-native-image-crop-picker/pull/292

PR #292, opened 2017-04-06
#283 Create DIRECTORY_PICTURES if it doesn't exist

andrevargas:
According to [official docs example](https://developer.android.com/reference/android/os/Environment.html#getExternalStoragePublicDirectory(java.lang.String)), you need to make sure that the external storage directory you want to manipulate (`Environment.DIRECTORY_PICTURES`, in this case) exists.

When running the previous code in Android 4.4 (KitKat), the method `PickerModule#createImageFile()`  throws an IOException, because there was no existing directory when calling `File#createTempFile()`. This behavior was documented in the issue #283.

So, I've just called `path.mkdirs()` to create the directory, inside an `if` statement that checks if the folder doesn't exist (which is not necessary, but make things more clear, though).


ivpusic:
tnx!

Issue #287, opened 2017-03-31
'React/RCTBridgeModule.h' file not found after project update


habovh:
### Version
Tell us which versions you are using: 

- react-native-image-crop-picker v0.12.9
- react-native v0.42.3

### Platform

- iOS

### Expected behaviour

Compilation should work.

### Actual behaviour

Compiling project fails with the following error:
```
/Users/Jordan/Repositories/ios-app/node_modules/react-native-image-crop-picker/ios/ImageCropPicker.h:17:9: 'React/RCTBridgeModule.h' file not found
```

Just as if Xcode was not able to access React header files.

### Attachments

I followed the install instructions in the README, using cocoa pods.
The compilation worked just fine before I updated to RN 42. I updated RNICP as well so it has the new import style that started from RN 40+.
Somehow this is still not working, despite all efforts.

alaingoldman:
Same

jakehasler:
@habovh and @alaingoldman:
I was able to fix this by updating "imageCropPicker.xcodeproj" > "Build Settings" > "Header Search Paths" from this:
```
$(SRCROOT)/../example/node_modules/react-native/React
$(SRCROOT)/../example/node_modules/react-native/Libraries/Image
```
To this:
```
$(SRCROOT)/../../react-native/React
$(SRCROOT)/../../react-native/Libraries/Image
```

Which I believe uses the React Version that is installed in your local project, instead of the one that is in the example project of the repo itself.

Hope this helps --

ivpusic:
fixed with https://github.com/ivpusic/react-native-image-crop-picker/issues/287

PR #304, opened 2017-04-17
Update HEADER_SEARCH_PATHS to be more compatible

jakehasler:
Referencing #287, the previous header search paths reference the React Native version from the example project, as opposed to that of the actual source project that the lib is included from. It was causing errors if there was inconsistency between the example and the actual project.


ivpusic:
tnx

PR #321, opened 2017-05-13
Add ratioX and ratioY options

inferusvv:
I've added `ratioX` and `ratioY` parameters. Also renamed `width` and `height` options to `maxWidth` and `maxHeight`

### DIFF
--- a/android/src/main/java/com/reactnative/ivpusic/imagepicker/PickerModule.java
+++ b/android/src/main/java/com/reactnative/ivpusic/imagepicker/PickerModule.java
@@ -74,14 +74,17 @@ class PickerModule extends ReactContextBaseJavaModule implements ActivityEventLi
     private ReadableMap options;
 
 
-    //Grey 800
+    // Grey 800
     private final String DEFAULT_TINT = "#424242";
     private String cropperTintColor = DEFAULT_TINT;
 
-    //Light Blue 500
+    // Light Blue 500
     private final String DEFAULT_WIDGET_COLOR = "#03A9F4";
     private int width = 200;
     private int height = 200;
+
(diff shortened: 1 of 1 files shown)


ivpusic:
tnx for PR. However, I will not merge it. First reason is because width/height should not be renamed. Second thing is that I think customizing aspect ratio is not a general requirement. I believe that you found use-case for this, so I recommend you to fork this repo and add that feature to your clone.

prashantsaini1:
@ivpusic thanks for this module first. However, I believe aspect-ratio is the first requirement than the width/height parameters. Any image can be of smaller resolution than the provided width/height values and that's where the aspect-ratio comes in. Most of the use-cases only take aspect-ratio into account rather than width/height. e.g. social media apps profile pics selection, etc.

karltaylor:
@ivpusic I am interested as to why you think "_customizing aspect ratio is not a general requiremen_` when this is what the `width` and `height` option does on iOS.

```
// Returns a custom rect for the mask.
- (CGRect)imageCropViewControllerCustomMaskRect:
(RSKImageCropViewController *)controller {
    CGSize maskSize = CGSizeMake(
                                 [[self.options objectForKey:@"width"] intValue],
                                 [[self.options objectForKey:@"height"] intValue]);

    CGFloat viewWidth = CGRectGetWidth(controller.view.frame);
    CGFloat viewHeight = CGRectGetHeight(controller.view.frame);

    CGRect maskRect = CGRectMake((viewWidth - maskSize.width) * 0.5f,
                                 (viewHeight - maskSize.height) * 0.5f,
                                 maskSize.width, maskSize.height);

    return maskRect;
}
```

I thought this would do the same thing.

PR #326, opened 2017-05-19
Fix getMimeType on Android. It's wrong if the name has space.

phuochau:
If the name has space or doesn't include the extension, `getMimeType` will return null.

### DIFF
--- a/android/src/main/java/com/reactnative/ivpusic/imagepicker/PickerModule.java
+++ b/android/src/main/java/com/reactnative/ivpusic/imagepicker/PickerModule.java
@@ -17,6 +17,8 @@ import android.support.v4.app.ActivityCompat;
 import android.support.v4.content.FileProvider;
 import android.util.Base64;
 import android.webkit.MimeTypeMap;
+import android.util.Log;
+import android.content.ContentResolver;
 
 import com.facebook.react.bridge.ActivityEventListener;
 import com.facebook.react.bridge.Callback;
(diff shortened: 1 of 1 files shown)


phuochau:
@ivpusic Pls help to review this PR and let me know if there are any issues. I just fixed and it's working well on my project :)

ivpusic:
could you explain to me what is happening here?

phuochau:
Hi @ivpusic , when the name has the space (e.g: Animals 14 - 4K res.mp4) or does not have ext (e.g: Animals 14 - 4k res), `getMimeType` will return null. It's not right.

We should check the mime type via `ContentResolver`.

phuochau:
Hi @ivpusic any chance to merge it into master?

superandrew213:
@ivpusic this looks good to me. Files with spaces in the file name don't work at the moment. For example `/location/of/my/cool file.mp4` won't work. `getMimeType` will return null. @phuochau PR fixes this.

See here: https://stackoverflow.com/a/31691791

This won't be an issue for media taken with the device but downloaded media that could have spaces in filename will return null for mime type.

android/src/main/java/com/reactnative/ivpusic/imagepicker/PickerModule.java
```diff
@@ -393,14 +396,19 @@ private String getBase64StringFromFile(String absoluteFilePath) {
...
+    private String getMimeType(String url) {
+      String mimeType = null;
+      Uri uri = Uri.fromFile(new File(url));
+      if (uri.getScheme().equals(ContentResolver.SCHEME_CONTENT)) {
+          ContentResolver cr = reactContext.getContentResolver();
+          mimeType = cr.getType(uri);
+      } else {
+          String fileExtension = MimeTypeMap.getFileExtensionFromUrl(uri
+                  .toString());
+          mimeType = MimeTypeMap.getSingleton().getMimeTypeFromExtension(
```
superandrew213: @phuochau can you check that fileExtension is not null here?

```
if (fileExtension != null) {
  mimeType = MimeTypeMap.getSingleton().getMimeTypeFromExtension(fileExtension);
}
```

PR #335, opened 2017-05-31
Add more color options in RN Image crop picker

sahil290791:
Now we can specify **Toolbar**, **Statusbar** and **active-widget** color. If not specified it defaults to default colors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,9 @@ ImagePicker.clean().then(() => {
 | height      | number      | Height of result image when used with `cropping` option |
 | multiple | bool (default false) | Enable or disable multiple image selection |
 | includeBase64 | bool (default false) | Enable or disable returning base64 data with image |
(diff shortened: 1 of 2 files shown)


ivpusic:
tnx for PR. I think there is one mistake in updated readme file. `cropperActiveWidgetColor` is missing

sahil290791:
@ivpusic I kept same name for `cropperActiveWidgetColor` => `cropperTintColor` as this if remove this option it will cause the problem to users who are already using this plugin.

ivpusic:
sure, but you are not reading `cropperTintColor` anymore (https://github.com/ivpusic/react-native-image-crop-picker/pull/335/files#diff-b63ce3ce1e5c477c1141d6db752515d6L115), so it should not be in the readme. I suggest to remove `cropperTintColor` from readme as it is not used anymore in the code (I will write down breaking change), and document `cropperActiveWidgetColor` in the readme file.

sahil290791:
Sure, will update the code.

sahil290791:
@ivpusic Have replaced `cropperTintColor` with `cropperActiveWidgetColor`.

ivpusic:
thanks

PR #345, opened 2017-06-08
Update PickerModule.java

GNarek:
As type of activeWidgetColor is int and type of DEFAULT_TINT is String it's causing error like: int cannot be dereferenced


ivpusic:
thanks

Issue #75, opened 2016-09-17
Some questions


npomfret:
Sorry, this isn't an issue - I just have some questions.
- is the EXIF metadata copied from the source image to the cropped result?
- is it possible to programatically access the crop function?

And 2 requests:
- could the exif data be added to the response?  
- could the original be added to the response also?

...I'd be happy to submit a PR if you're open to the ideas

ivpusic:
1. no
2. no

Requests:
1. Open for PR
2. If you mean as base64, it would be too slow

npomfret:
"could the original be added to the response also?" - i didn't mean base64 data, just a reference to the original image.

ivpusic:
hmm, then maybe we could do it in a way that you add metatada key to the response with all metadata you discovered?

npomfret:
So maybe add a lump to the response called 'original' or something, and in it include the image data and all the exif meta data?

ivpusic:
still didn't get you what you mean by "original". Currently in iOS is hard or impossible to get original image url which will work later on upload. That is the reason why I decided to save NSData to tmp location and return tmp path to the user. My idea is to attach "metadata" key along with existing response keys, and "metadata" key would be key/value pairs of extracted exif data

npomfret:
By "original" I just mean the data that would have been returned if the user had just picked an image from the Camera Roll.  This is the image that will have all the meta data in tact.

ivpusic:
as you can see we are already returning a lot of this data to the user. It would be great if you could create key/value `metadata` object in response, and fill it with exif and missing location data (if available). This would be actually very cool

npomfret:
Adding the original asset URI to the existing response would be enough I think.  It could be used to look up the original object in the camera roll and from there you could inspect the original for its metadata. 

Maybe thats a good start?

ivpusic:
hm, in my opinion would be much easier for the users if we extract data automatically without this extra step. How can we from react native look at the original image metadata (if we have url)?

npomfret:
There are APIs for getting the meta data out, but you need the original url first.  Can you add this in?  I'm able to get the example app working so don't feel comfortable submitting a PR that I can't even test.  

You could add metadata to this project, but after some though I'm not sure it belongs here.  But without the original URL, you can't get anything.

Looks like this code is very similar to react-native-image-picker which already returns the original url in the response.  Could just borrow the code from them, I'm sure its just a couple of lines.

[...]

PR #349, opened 2017-06-15
Exif data extraction

nico1510:
Hi @ivpusic this is a pull request which adds exif data to the image response which some people (including me) asked for here https://github.com/ivpusic/react-native-image-crop-picker/issues/75.
Everything works for me but I'm a Java/JS dev so you should check the objective C code closely. Unfortunately I don't own an iPhone so I couldn't test if it also works when taking a picture with the camera (https://github.com/ivpusic/react-native-image-crop-picker/compare/master...nico1510:master#diff-c227004bfb2fcf3e01a87ae371a37a2dR159)

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,7 @@ ImagePicker.clean().then(() => {
 | height      | number      | Height of result image when used with `cropping` option |
 | multiple | bool (default false) | Enable or disable multiple image selection |
 | includeBase64 | bool (default false) | Enable or disable returning base64 data with image |
+| includeExif | bool (default false) | Enable or disable returning exif data with image. Does not work when cropping is set to `true`. |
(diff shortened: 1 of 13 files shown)


pho3nixf1re:
Why was this PR closed? This seems to be an unresolved issue that is working in your fork.

nico1510:
Even though this pull request was closed the code was merged at some point into master I think.

pho3nixf1re:
I see that now. We are using a rather old version of this tool before this was merged in. Time to update a dependency!

dancherb:
Exif data is working for me on Android but not on iOS, more info here: https://github.com/ivpusic/react-native-image-crop-picker/issues/751 - please could you have a look?

PR #361, opened 2017-06-26
Removed TeamID from XCode-project-file.

itinance:
This PR removes your TeamID out of the XCode-Projectfiles. When the TeamID is included, XCode 
does it not allow to create an Archive unless the current developer is member of your team.

The issue that will be fixed here is:

> No "iOS Development" signing certificate matching team ID "5WFBC66J2G" with a private key was found.

More information about this issue you can find here: https://github.com/ruslanskorb/RSKImageCropper/issues/164

It does not break any compatibility.


ivpusic:
thanks

PR #369, opened 2017-07-10
Fix RSKImageCropper not found

magrinj:
When I try to compile my app in release I got a compilation error with this line:

```
#import "RSKImageCropper/RSKImageCropper.h"
```

I try everything, clean the cache, remove pod dependencies and reinstall, unlink an link again, but nothing works, the only solution I found is to change the import to this:

```
#import <RSKImageCropper/RSKImageCropper.h>
```

Doesn't seems to be a big difference, but it's now working for me


ivpusic:
tnx

magrinj:
You're welcome :)

PR #308, opened 2017-04-22
Add podspec

johncblandii:
FYI @ivpusic. I'll leave the documentation updates to you.

Fixes #307 

Now you simply add this to your `Podfile` with no reference to any other dependency and run your app:
```
pod 'react-native-image-crop-picker', path: 'node_modules/react-native-image-crop-picker'
``` 

Note: I put my `Podfile` in the app root (not `[root]/ios`) so the path would change if you used default pattern of sticking it in `[root]/ios`.
```
pod 'react-native-image-crop-picker', path: '../node_modules/react-native-image-crop-picker'
```

PR #372, opened 2017-07-11
Added podspec

poyannabati:
Using #308 as a reference, I added a working podspec once more. 

@ivpusic; I saw you mentioned that people were confused on how to use it. I think we could reorganize the docs to make it clearer. The docs need to be updated on how to use this as well. Let me know if you want me to do this.

### DIFF
--- a/ios/ImageCropPicker.h
+++ b/ios/ImageCropPicker.h
@@ -22,7 +22,7 @@
 #import "QBImagePicker.h"
 #import "RSKImageCropper.h"
 #else
-#import "QBImagePicker/QBImagePicker.h"
+#import "QBImagePickerController/QBImagePickerController.h"
 #import <RSKImageCropper/RSKImageCropper.h>
 #endif
 
--- a/react-native-image-crop-picker.podspec
+++ b/react-native-image-crop-picker.podspec
@@ -1,0 +1,18 @@
+require "json"
+package = JSON.parse(File.read('package.json'))
+
+Pod::Spec.new do |s|
+  s.name          = package['name']
+  s.version       = package['version']
+  s.summary       = package['description']
+  s.author        = "Ivan Pusic"
+  s.license       = package['license']
+  s.requires_arc  = true
+  s.homepage      = "https://github.com/ivpusic/react-native-image-crop-picker"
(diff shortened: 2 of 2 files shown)


ivpusic:
I guess this wont work now for people who dont use cococaposa because you change header import.

ios/ImageCropPicker.h
```diff
@@ -22,7 +22,7 @@
 #import "QBImagePicker.h"
 #import "RSKImageCropper.h"
 #else
```
chaitanya0bhagvan: To handle cases where this library is not installed using cocoapods we should include another conditon

```
#elseif __has_include("QBImagePicker/QBImagePicker.h")
#import "QBImagePicker/QBImagePicker.h"
#import <RSKImageCropper/RSKImageCropper.h>

```

ivpusic:
A lot of people were having issues with podspec, so I removed it completely and added installation instructions how to install a library without it.

chaitanya0bhagvan:
Do you see any issue in retaining the podspec file and the changes to the
header file?

ivpusic:
yes...it was breaking the installation. refer to https://github.com/ivpusic/react-native-image-crop-picker/issues/421 and https://github.com/ivpusic/react-native-image-crop-picker/issues/414.

I tried it by myself and was also getting compilation issues.

chaitanya0bhagvan:
I was looking at those issues and I realised the read me I created assumed
that users who would use podspec had installed react native using podspec
too.

I should have been very clear that users should link this library using
react-native link if they are not using podspec to install react native.

You can confirm that retaining the header file changes and podspec won't
cause any build issues if you link this library using react-native link.

Podspec is for those developers who use pods to manage all their dependency.


On 25-Aug-2017 2:04 AM, "Ivan Pusic" <notifications@github.com> wrote:

yes...it was breaking the installation. refer to #421
<https://github.com/ivpusic/react-native-image-crop-picker/issues/421> and
#414 <https://github.com/ivpusic/react-native-image-crop-picker/issues/414>.

I tried it by myself and was also getting compilation issues.

—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub
<https://github.com/ivpusic/react-native-image-crop-picker/pull/372#issuecomment-324749683>,
or mute the thread

PR #373, opened 2017-07-12
iOS asset-url added

havenS:
Added the `asset-url` for ios, allow to use package like `react-native-exif` to later retrieve EXIF data (geolocation, datetime, various info)


ivpusic:
closing this as probably not relevant anymore. Recently exif support is added directly into this library. Also library is returning additional fields, like localIdentifier, sourceURL.

Issue #353, opened 2017-06-21
Error when adding pictures from Camera on Android


puelocesar:
### Version
Tell us which versions you are using: 

- react-native-image-crop-picker v0.13.1
- react-native v0.44.2

### Platform
- Android

### Expected behaviour
Getting correct photo after taking it

### Actual behaviour
Getting error when taking picture from camera on Android: 
Error: Attempt to invoke virtual method 'int java.util.concurrent.atomic.AtomicInteger.addAndGet(int)' on a null object reference

### Steps to reproduce

1. ImagePicker.openCamera(this.cameraOptions).then(image => {
            console.log(image)
        })

2. take a picture, clicks OK

3. getting yellow box error, and console.log(image) never gets called

ps: it works fine on iOS

### Attachments

// stacktrace or any other useful debug info
Error: Attempt to invoke virtual method 'int java.util.concurrent.atomic.AtomicInteger.addAndGet(int)' on a null object reference
Error: Attempt to invoke virtual method 'int java.util.concurrent.atomic.AtomicInteger.addAndGet(int)' on a null object reference
    at createErrorFromErrorData (http://localhost:8081/index.android.bundle?platform=android&dev=true&hot=false&minify=false:6722:15)
    at http://localhost:8081/index.android.bundle?platform=android&dev=true&hot=false&minify=false:6680:25
    at MessageQueue.__invokeCallback (http://localhost:8081/index.android.bundle?platform=android&dev=true&hot=false&minify=false:7001:16)
    at http://localhost:8081/index.android.bundle?platform=android&dev=true&hot=false&minify=false:6847:16
    at MessageQueue.__guard (http://localhost:8081/index.android.bundle?platform=android&dev=true&hot=false&minify=false:6936:9)
    at MessageQueue.invokeCallbackAndReturnFlushedQueue (http://localhost:8081/index.android.bundle?platform=android&dev=true&hot=false&minify=false:6846:12)
    at http://localhost:8081/debuggerWorker.js:71:58

deathemperor:
+1. Having the same issue

ivpusic:
do you have some time to investigate this?

deathemperor:
I think I know why. it's because I set "multiple: true" for openCamera. Removing that solved it.

arribbar:
It works for me, thank you @deathemperor 
Can you make a PR ? Or do you prefer I do it ?

deathemperor:
[...]

PR #380, opened 2017-07-20
fixed a bug when using openCamera with [multiple: true]

deathemperor:
related issue: https://github.com/ivpusic/react-native-image-crop-picker/issues/353


ivpusic:
thanks!

PR #381, opened 2017-07-20
[iOS] Fix  a bug that cause a {0,0} size image

Howard-Wang-Hao:
-[origin] if ([maxWidth integerValue] == 0 || [maxHeight integerValue] == 0) {
-When pick a width< height image and only set "compressImageMaxWidth",will cause  {0,0}size image.
-Now i fix it.


ivpusic:
thanks a lot!

Issue #384, opened 2017-07-22
How to get permanent filepath and avoid copying assets by returning Asset-Library://


gameboyVito:
I am using the multiple selection. I would like to know is there any way to make the tmp file to be permanent, because I want to use these selected photos in the future, at least after users restart the app. If you can just return the tag of images inside CameraRoll, like `Asset-Library://..........`, that would be nice!

gameboyVito:
I have read your code, it seems like it always executes the `persistFile` method, even if I just only use the multiple-selection, instead of cropping. May be you should see if the `cropping ` property is enabled. If it is disabled, then we shouldn't create any temp files. It is unnecessary and time-consuming, if you are trying to select 1000 photos or 100 videos and without using compression feature. 

In order words, if the user still wants to get a read-write file, they can use `react-native-fs` or `react-native-fetch-blob` to easily create a temp file.

gameboyVito:
Another solution is: simply create a new method call: `getCacheDirectory`, the cacheDir is not the same as tmpDir because it will not be purged by the system automatically.

```
- (NSString*) getCacheDirectory {
    NSString *TMP_DIRECTORY = @"react-native-image-crop-picker/";
    NSString *cacheDir = [NSSearchPathForDirectoriesInDomains(NSCachesDirectory, NSUserDomainMask, YES) lastObject];
    NSString *tmpFullPath = [cacheDir stringByAppendingString:TMP_DIRECTORY];

    BOOL isDir;
    BOOL exists = [[NSFileManager defaultManager] fileExistsAtPath:tmpFullPath isDirectory:&isDir];
    if (!exists) {
        [[NSFileManager defaultManager] createDirectoryAtPath: tmpFullPath
                                  withIntermediateDirectories:YES attributes:nil error:nil];
    }

    return tmpFullPath;
}
```

Then, added a new bool property called `cached`, to set whether to use `tmp` or `cache`, if using cached, developers should manually call `ImagePicker.clean()`. It would be the quickest solution, instead of re-writing the main logic.

ivpusic:
the reason for tmp files is because if reading directly from original file path you will get permission error. Creating tmp files where app has access is a workaround to make it possible to read files from react-native app. If you know solution to this problem, please let me know.

gameboyVito:
[...]

PR #385, opened 2017-07-25
returned localIdentifier and filename in response

gameboyVito:
Reference on #384 

This PR adds two new response property: one is `localIdentifier`, the other is `filename`

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,41 +71,43 @@ ImagePicker.clean().then(() => {
 
 #### Request Object
 
-| Property        | Type           | Description  |
-| ------------- |:-------------:| :-----|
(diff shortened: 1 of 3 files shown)


ivpusic:
thanks a lot!

gameboyVito:
You are welcome

kholiavko-roman:
Guys, now when I take photo from camera it does not return filename, how to get it ?

ivpusic:
@gameboyVito could you help here?

kholiavko-roman:
Now I see, that you edited readme, and filename was marked as ios only.  but how I can get filename on android ?

ivpusic:
should be returned as part of url

PR #386, opened 2017-07-25
Fix for adding files more than 30 at once for iOS

ZmicierBiesau:
I have getting crashes with selecting >30 images at once some times on iOS. I noticed that the less free memory you have, the more chance to get the crash you have.


ivpusic:
thanks. maybe we need the same for video block?

ZmicierBiesau:
Just checked video. Looks like all is ok with video. I think with images problems were caused by UIImage objects memory managing and there is no such problems with video, because video is processing in another way.

ivpusic:
thanks a lot for help!

ZmicierBiesau:
you are welcome

Issue #388, opened 2017-07-26
New feature and performance enhancement


gameboyVito:
Reference Post #384 #385 
Dear author @ivpusic ,

We have discussed some new features listed below,
1. Added new properties localIdentifier and filename to be returned
2. Check if cropping and compression are both disabled. If yes, then do not return tmp path and generate copied files but only return localIdentifier. It is the essential to improve the performance when users want to select thousands of photos in the multiple selection mode.

The first one has been merged #385 
However, if someone don't know how to use the localIdentifier to get the `assetURL`. Here is an Example code:
```
/**
 * I write a litte util method to convert localIdentifier to assetURL in JavaScript
 * @param localIdentifier looks like 91B1C271-C617-49CE-A074-E391BA7F843F/L0/001
 * @param ext the extension: JPG, PNG, MOV
 * @returns {string}
 */
export const convertLocalIdentifierToAssetLibrary = (localIdentifier, ext) => {
    const hash =  localIdentifier.split('/')[0];
    return `assets-library://asset/asset.${ext}?id=${hash}&ext=${ext}`;
};
```

And the second one I would like to make a new PR.
Sent. #389

ivpusic:
thanks. Isn't AlAssetLibrary and assets-library deprecated?

gameboyVito:
No, the react-native `<Image />` component is still relying on this library. AlAssetLibrary is just not suggested to use after iOS8, but both AlAssetLibrary and PhotoKit are still available until now.

PR #389, opened 2017-07-26
Multiple selection performance enhancement

gameboyVito:
reference on #388 

This PR boost the response speed when we do not want to use cropping, compression functionalities. The tricky part is to skip compression, and skip generating tmp files, this feature will save tons of time. With the help of localIdentifier(the id of each photo asset), we can always get what we want.

Suitable conditions:
For image:
1. `multiple` is true,
2. `compressImageMaxWidth` is not set
3. `compressImageMaxHeight` is not set
4. `compressImageQuality` is 1 or not set

For video:
1. `compressVideoPreset` is set to be `Original`

### DIFF
--- a/ios/ImageCropPicker.m
+++ b/ios/ImageCropPicker.m
@@ -69,6 +69,24 @@ RCT_EXPORT_MODULE();
     return self;
 }
 
++ (NSString*)contentTypeForImageData:(NSData *)data {
+    uint8_t c;
+    [data getBytes:&c length:1];
+    
+    switch (c) {
+        case 0xFF:
+            return @"image/jpeg";
+        case 0x89:
+            return @"image/png";
+        case 0x47:
+            return @"image/gif";
+        case 0x49:
+        case 0x4D:
+            return @"image/tiff";
+    }
+    return @"";
+}
+
 - (void (^ __nullable)(void))waitAnimationEnd:(void (^ __nullable)(void))completion {
     if ([[self.options objectForKey:@"waitAnimationEnd"] boolValue]) {
         return completion;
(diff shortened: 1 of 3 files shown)


ivpusic:
hmm, but in this case how user can display image from react-native app if path is not returned?

gameboyVito:
https://github.com/ivpusic/react-native-image-crop-picker/issues/388

Using this can help to get an assetURL.
```
/**
 * I write a litte util method to convert localIdentifier to assetURL in JavaScript
 * @param localIdentifier looks like 91B1C271-C617-49CE-A074-E391BA7F843F/L0/001
 * @param ext the extension: JPG, PNG, MOV
 * @returns {string}
 */
export const convertLocalIdentifierToAssetLibrary = (localIdentifier, ext) => {
    const hash =  localIdentifier.split('/')[0];
    return `assets-library://asset/asset.${ext}?id=${hash}&ext=${ext}`;
};
```

gameboyVito:
You can always use `<Image source={{uri: assetURL}}/>` to display the content. It will not be clean by system just like tmp files, instead, the assetURL is the id of each image inside CameraRoll. Might be you want to write this func into native side.

gameboyVito:
The Image component written by react-native is still using AlAssetLibrary. However, I found the localIdentify can be easily converted to assetURL. And wrap it into <Image/> component, it can be displayed normally. But if you put a tmp path into the <Image/> component, after a few hours or days, these files will be cleaned and no way to find it back.

gameboyVito:
In some conditions, what if users do not want to use cropping and compression, but only want to select some photos putting them into a Favourites page, or picking some pending photos to be uploaded in the future?

ivpusic:
Thanks. I see your points. However, I would suggest returning path in all cases. The reason is consistency and convenience. On Android side path will be always returned, so users can assume that they can safely put the path into Image component, and the image will be shown. On iOS side if we would go this way, the user would have to check if the path is returned, and if not, check platform and if the local identifier is present, and then call a function to format final path, which is really a lot of additional work which is not necessary. 

Therefore I suggest to format and return path which points to the original image (path field in response), and the user can again assume that it exists, it is valid, and it can be used for uploads or for showing the image in the app.

What do you think?

ios/ImageCropPicker.m
```diff
@@ -363,29 +381,45 @@ - (void) getVideoAsset:(PHAsset*)forAsset completion:(void (^)(NSDictionary* ima
...
+         if ([self.options[@"compressVideoPreset"]  isEqual: @"Original"]) {
+             //if the compress video preset is set to be original, then do not compress the video
+             NSLog(@"video info: %@", info);
+             AVAssetTrack *track = [[asset tracksWithMediaType:AVMediaTypeVideo] firstObject];
+             completion([self createAttachmentResponse:@""
+                                   withLocalIdentifier:forAsset.localIdentifier
+                                          withFilename:sourceURL.lastPathComponent
+                                             withWidth:[NSNumber numberWithFloat:track.naturalSize.width]
+                                            withHeight:[NSNumber numberWithFloat:track.naturalSize.height]
+                                              withMime:@"video/quicktime"
```
ivpusic: I think we cannot hardcode mime type here, because not always returned video is quicktime. Video can also be mp4 if it is sent via whats app for example. Is there some other way to find a mime type?

ios/ImageCropPicker.m
```diff
@@ -458,31 +492,46 @@ - (void)qb_imagePickerController:
...
-                                     }]];
-                                     return;
+                                 if ([self.options[@"compressImageMaxWidth"] integerValue] == 0 && [self.options[@"compressImageMaxHeight"] integerValue] == 0 && [self.options[@"compressImageQuality"] isEqualToNumber:@1]) {
+                                     
+                                     [selections addObject:[self createAttachmentResponse:@""
+                                                                      withLocalIdentifier:phAsset.localIdentifier
+                                                                             withFilename:sourceURL.lastPathComponent
+                                                                                withWidth:[NSNumber numberWithFloat:imageT.size.width]
+                                                                               withHeight:[NSNumber numberWithFloat:imageT.size.height]
+                                                                                 withMime:[ImageCropPicker contentTypeForImageData:imageData]
```
ivpusic: is there a way to avoid our custom function to get a mime type?

ivpusic:
also, would it be possible to introduce the same features when `multiple` is false? I think for single images/videos this feature should be also included, to have consistency.

gameboyVito:
Yeah, I haven't implemented this feature to single image, but single video selection is done. In my mind, I would suggest you add a new option name called `noCopy`, or some other better name you can imagine. If `noCopy` is ture, then the path will return the assertURL. Otherwise, it will still return the tmp filepath. Both of these assertURL and tmpPath will be returned as one property: `path`, so it will not interfere to people who have been using this library already. And the android side will still remain the same and has no impact. 

Thus, if you agree with this, I can implement this feature to single image and find a way to get the correct mime type for videos. 

As far as I know, like the ios assertURL, the android has a similar unique URL started with prefix `content://`. We can easily retrieve the original path and the binary data from this kind of contentURL. See https://stackoverflow.com/questions/5657411/android-getting-a-file-uri-from-a-content-uri. Therefore, after this ios part is done, android will be similar. Either contentURL or tmpPath will be returned inside property `path`.

ios/ImageCropPicker.m
gameboyVito: I think the `MobileCoreServices` can do the trick. See https://stackoverflow.com/questions/1363813/how-can-you-read-a-files-mime-type-in-objective-c
```
/**
 ** the path should be the sourceURL, instead of the tmp filepath
 **/

+ (NSString*) mimeTypeForFileAtPath: (NSString *) path {
    if (![[NSFileManager defaultManager] fileExistsAtPath:path]) {
        return nil;
    }
    CFStringRef UTI = UTTypeCreatePreferredIdentifierForTag(kUTTagClassFilenameExtension, (CFStringRef)[path pathExtension], NULL);
    CFStringRef mimeType = UTTypeCopyPreferredTagWithClass (UTI, kUTTagClassMIMEType);
    CFRelease(UTI);
    if (!mimeType) {
        return @"application/octet-stream";
    }
    return (__bridge NSString *)(mimeType);
}
```

ios/ImageCropPicker.m
gameboyVito: See below.

nico1510:
@gameboyVito thanks for your efforts here I think avoiding temp file creation is absolutely necessary once the number of selected images gets above ~20. Is there a followup PR to this one or did you encounter problems which can't be solved ?

nico1510:
also @gameboyVito you wrote
> You can always use `<Image source={{uri: assetURL}}/>` to display the content. 

this doesn't work for me though.. at least not on the simulator. I used your 'hack' to convert the localIdentifier returned by this library like follows:
```javascript
<Image source={{uri: convertLocalIdentifierToAssetLibrary(localIdentifier, 'JPG')}}/>
```

but I get the error: 
```
No suitable image URL loader found for assets-library://asset/asset.JPG?id=213D568C-FEA3-441B-A3C8-BE9A9B5CAC39&ext=JPG
```
@ivpusic did this work for you or haven't you tried yet ?

ios/ImageCropPicker.m
Louies89: ImagePicker.openPicker({
      mediaType:'photo'  <==Can be 'video'or 'any'
    })
    .then(file => {
      console.log(file.mime);  <== This shall give mime type of the media chosen
    })

Issue #395, opened 2017-07-31
Feature Request - Should provide minFiles options to restrict the minimum selection


gameboyVito:
I believe this image picker should provide `minFiles` options to restrict the minimum file selection. As the QBImagePicker has already provided a method called `isMinimumSelectionLimitFulfilled` which will check if the minimum condition is fulfilled or not. If not, then the done button will be disabled, otherwise, it will be enabled.

I would like to add a new option called `minFiles` to openPicker API. Do you agree with this? If yes, I am happy to send you a PR. @ivpusic

ivpusic:
yes, awesome.

PR #407, opened 2017-08-04
added minFiles option, and a new response called sourceURL

gameboyVito:
reference #395 

- added a new option: `minFiles`
- added a new response property: `sourceURL`
- updated README

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,7 @@ ImagePicker.clean().then(() => {
 | cropperStatusBarColor (android only)    |        string (default `#424242`)        | When cropping image, determines the color of StatusBar. |
 | cropperToolbarColor (android only)      |        string (default `#424242`)        | When cropping image, determines the color of Toolbar. |
 | cropperCircleOverlay                    |           bool (default false)           | Enable or disable circular cropping mask. |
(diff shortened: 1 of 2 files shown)


ivpusic:
thanks a lot!

ivpusic:
just one question...do we need to add something like https://github.com/ivpusic/react-native-image-crop-picker/blob/master/ios/ImageCropPicker.m#L399. Bug was reported some days ago because in case if key value was empty, app was crashing. Reference https://github.com/ivpusic/react-native-image-crop-picker/pull/397

gameboyVito:
I guess it is because the app is running below ios8, which does not support PhotoKit. Yes, I think it is needed to check whether it is empty.

ivpusic:
do you have 5min to add quick push?

gameboyVito:
Sure

gameboyVito:
I think this #397 has already fixed this issue.

ivpusic:
but not for newly added `sourceURL` property. Or this one cannot be empty?

gameboyVito:
Okay, I get it. Fixed - #408

PR #441, opened 2017-09-11
Added video docs

fendorio:
Added additional example to the code snippets provided, showing how to select videos with the ImagePicker. 

Also documented the behaviour whereby having the prop 'cropping' set to true results in no videos being shown on Android. I have stated that this issue is known to have happened, rather than stating that it happens all the time, due to the relative poor test coverage across the broad spectrum of Android devices available.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,18 @@ ImagePicker.openPicker({
 });
 ```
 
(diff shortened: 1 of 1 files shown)


ivpusic:
Thanks!

PR #448, opened 2017-09-18
Differentiate between video and pictures

sinlyly:
Differentiate between video and pictures


ivpusic:
I would not implement this on native side. People can do this from react-native side to customize it for their projects.

Issue #431, opened 2017-08-30
undefined is not an objdect (evaluating 'c.default.openPicker') 


congdufs:
### Version
Tell us which versions you are using: 

- react-native-image-crop-picker v0.?.?
- react-native v0.?.?

### Platform
Tell us to which platform this issue is related

- iOS
- Android

### Steps to reproduce

1.

2.

3.

### Attachments

// stacktrace or any other useful debug info

dgurns:
Getting same thing here on iOS (working fine on Android)

RN 0.44
react-native-image-crop-picker 0.16.0

![simulator screen shot sep 1 2017 1 12 07 pm](https://user-images.githubusercontent.com/1173791/29980292-33609f9c-8f17-11e7-99bd-d1b7c8b4b80a.png)

ivpusic:
installation of library failed. Please read installation instructions again, follow each step and make sure each step succeeds.

dgurns:
I did follow the steps exactly and used CocoaPods. It worked fine while running on my iPhone 6 via Xcode. However, on the Simulator it did not work. It failed right before the permissions step.. maybe it's something to do with that?

anjalsaneen123:
@dgurns did you got what is the solution, i'm facing same error?

Mandeep03:
Add this in you MainApplication.java, sometimes react-native link does not work perfectly.
<img width="532" alt="screen shot 2017-09-07 at 7 16 56 pm" src="https://user-images.githubusercontent.com/7822632/30166555-4cc19d8a-9401-11e7-98e9-8000f76b87cd.png">

ropc:
I was having a similar issue on iOS. I also use other cocoapods dependencies so I made a podspec that works for me. Made a PR for it here: #454

ivpusic:
@congdufs please try suggestion from @Mandeep03

PR #454, opened 2017-09-20
Adds podspec for installation via cocoapods

ropc:
Fixes issue where `NativeModules.ImageCropPicker` (default export) would be undefined when installing with cocoapods. Possibly related to #431.

All this does is add a podspec for the iOS native module that you can reference from the rn app's `ios/Podfile`. Needed to change an import to make it work.

To install you should run

```
npm i react-native-image-crop-picker --save
```

and then in your `ios/Podfile` add:

```ruby
pod 'ImageCropPicker', :path => '../node_modules/react-native-image-crop-picker/ios'
```

The above should also go in the README.md.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -14,8 +14,14 @@ android {
     }
 }
 
+repositories {
+    maven {
+        url "$rootDir/../node_modules/react-native/android"
+    }
+}
+
 dependencies {
-    compile 'com.facebook.react:react-native:+'
+    compile 'com.facebook.react:react-native:0.48.1'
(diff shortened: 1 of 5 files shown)


anyexinglu:
Seems that ivpusic has [removed the podspec](https://github.com/ivpusic/react-native-image-crop-picker/commit/07f8763c977b7482649aadf75c417c56fa347043).

And in the README of the current version 0.17.3, ios/podfile should like this:
`platform :ios, '8.0'

target '<your_project_name>' do
    pod 'RSKImageCropper'
    pod 'QBImagePickerController'
end`

So should we adjust the ios/profile following your steps above?

PR #460, opened 2017-09-24
Update index.d.ts to mute the error message

freedomofme:
Try to mute the warning of "[ts] Module react-native-image-corp-picker has no default export"

After careful consideration, I have not increased the typescript example in the readme file, because the code of typescript and javascript are almost the same.

Really appreciate all your kind help.


ivpusic:
thanks

PR #497, opened 2017-11-01
TypeScript 2.6.0 default export fix

svbutko:
Added additional interface which grabs all functions and also added a constant which will be used as a default export


ivpusic:
thanks!

PR #521, opened 2017-11-26
[Android] Add option to set the cropper toolbar title

Fuhrmann:
I needed to change the toolbar title but didn't know any other way so I thought about adding this functionality. Currently, as my knowledge is limited to java, this is currently only available for Android.

Example usage:

```
ImagePicker.openCropper({
    path: 'image.png',
    width: 800,
    height: 800,
    cropperToolbarTitle: 'Please, crop your photo',
})
```

Will result in this:

image

If no title is provided, the default one from the library will be used.


ivpusic:
thanks

Issue #514, opened 2017-11-20
suggest:  android multiple 


giantss:
At present, many people reflect that android cannot be selected on many manufacturers. Can you please find a way to write a generic android multiple selector? Please

ivpusic:
currently, I don't have a time for this. If anyone from the community is having some time to do this, PR would be very welcome.

superandrew213:
@ivpusic can we have a TO DO list in the readme? I think 2 key features should be:

1. Multiple select for Android
2. Remote media for Android

ivpusic:
@superandrew213 good idea. Could you add it to the readme file?

PR #537, opened 2017-12-06
Add to do

superandrew213:
https://github.com/ivpusic/react-native-image-crop-picker/issues/514#issuecomment-345849475


ivpusic:
hmm, `Multiple select` is already present, but it depends on android version. More correct description would be to say `Standardize multiple select`.

superandrew213:
Done!

ivpusic:
thanks!

PR #552, opened 2017-12-20
Added all Smart Album types supported by iOS 9.0+

jasongaare:
I added some more Smart Album types, limited to those supported by iOS 9.0+

I've seen in the issues that it is believed these are limited by the `QBImagePicker` library, but from my testing that is not the case. Any `phassetcollectionsubtype` should work out of the box with that library, as far as I can tell. You may test as necessary.

Others available are:
- smartAlbumDepthEffect (iOS 10.2+)
- smartAlbumLivePhotos (iOS 10.3+)
- smartAlbumAnimated (iOS 11.0+)
- smartAlbumLongExposures (iOS 11.0+)

These could be added as well, but I was concerned about compatibility for some developers. If you'd like them included, I can add and resubmit.


jasongaare:
Also, I fixed a typo where the response key for `creationDate` was actually `creationDate:` (with the `:` at the end)

ivpusic:
already added with https://github.com/ivpusic/react-native-image-crop-picker/commit/6465d4e0bf1e38ba68cca0b65412999e009e72c1

Issue #449, opened 2017-09-18
when compressImageQuality = 1, image size is bigger than original size


yiky84119:
### Version
Tell us which versions you are using: 

- react-native-image-crop-picker v0.16.1
- react-native v0.47.1

### Platform
Tell us to which platform this issue is related

- iOS

### Expected behaviour

default width and height, set compressImageQuality to 1, get compressed image's size is bigger than original size. should return default NSData when set the default option.

### Actual behaviour

default width and height, set compressImageQuality to 1, get compressed image's size is bigger than original size. Almost increased by 300%.

### Steps to reproduce

1.

2.

3.

### Attachments

// stacktrace or any other useful debug info

ivpusic:
do you have some time to try to fix this and submit a PR?

fubar:
A change to skip compression in this scenario (#554) will go out with the next release. If you want to retrieve the original image without modification I'd suggest omitting all compressImage* parameters.

ivpusic:
fixed by https://github.com/ivpusic/react-native-image-crop-picker/pull/736

Issue #252, opened 2017-02-25
Compressed image is larger than original


cwRichardKim:
### Version
Tell us which versions you are using:

- react-native-image-crop-picker v0.11.2
- react-native v0.37.0

### Platform
Tell us to which platform this issue is related

- iOS

### Expected behaviour

Compressed image is smaller, or at the very least the same size as the original

### Actual behaviour

original (66kb): 
![1488041806216](https://cloud.githubusercontent.com/assets/5778691/23333224/f113f170-fb55-11e6-921e-7e31b0862b8d.jpg)

just size compression (433kb): 
![1488043261799](https://cloud.githubusercontent.com/assets/5778691/23333226/02794c26-fb56-11e6-8fc2-696880f877ac.jpg)

size and image compression (92kb): 
![14711143_10154622892393064_3707145447315550417_o](https://cloud.githubusercontent.com/assets/5778691/23333240/302d3042-fb56-11e6-9abd-9ec9849be264.jpg)

### Steps to reproduce

```
ImagePicker.openPicker({
        width: 700,
        height: 700,
        cropping: true,
        compressImageMaxHeight: 700,
        compressImageMaxWidth: 700,
        compressImageQuality: 0.6, // <- removed this for 2nd photo)
      }).then(image => {
        this._uploadPhotoToFirebase(image, index);
      }).catch(error => {
        console.log(error);
      });
```

### Attachments

// stacktrace or any other useful debug info

fubar:
A change to skip compression in this scenario (#554) will go out with the next release. If you want to retrieve the original image without modification I'd suggest omitting all compressImage* parameters.

superandrew213:
@ivpusic this can be closed.

ivpusic:
tnx @superandrew213

ivpusic:
fixed by https://github.com/ivpusic/react-native-image-crop-picker/pull/736

PR #554, opened 2017-12-23
Skip image compression when not desired (iOS)

fubar:
What do you guys think about this?

As reported in https://github.com/ivpusic/react-native-image-crop-picker/issues/252 and https://github.com/ivpusic/react-native-image-crop-picker/issues/449, file sizes end up around 3x larger when not specifying compression and resizing. Plus I need the ability to retrieve photos without any modifications.

This change skips the image compressor when the requested compression is 1 (or not specified) and max width/height are equal/larger than the original dimensions (or not specified). iOS only.

Feedback appreciated.

### DIFF
--- a/ios/ImageCropPicker.m
+++ b/ios/ImageCropPicker.m
@@ -429,6 +429,25 @@ RCT_EXPORT_METHOD(openCropper:(NSDictionary *)options
              };
 }
 
+// See https://stackoverflow.com/questions/4147311/finding-image-type-from-nsdata-or-uiimage
+- (NSString *)determineMimeTypeFromImageData:(NSData *)data {
+    uint8_t c;
+    [data getBytes:&c length:1];
+
+    switch (c) {
+        case 0xFF:
+            return @"image/jpeg";
+        case 0x89:
+            return @"image/png";
+        case 0x47:
+            return @"image/gif";
+        case 0x49:
+        case 0x4D:
+            return @"image/tiff";
+    }
+    return @"";
+}
+
 - (void)qb_imagePickerController:
 (QBImagePickerController *)imagePickerController
           didFinishPickingAssets:(NSArray *)assets {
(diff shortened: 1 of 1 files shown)


ios/ImageCropPicker.m
```diff
@@ -429,6 +429,25 @@ - (NSDictionary*) createAttachmentResponse:(NSString*)filePath withExif:(NSDicti
              };
 }
 
+// See https://stackoverflow.com/questions/4147311/finding-image-type-from-nsdata-or-uiimage
+- (NSString *)determineMimeTypeFromImageData:(NSData *)data {
```
fubar: I am aware of this comment: https://github.com/ivpusic/react-native-image-crop-picker/pull/389/files#r129528647

I'd still suggest this solution for performance reasons as it seems to be very reliable mechanism.

fazeelzama20: @fubar Hello, forgive if I am asking silly question here, if we have to implement this, what changes we have to do in our code. I have implemented the library and using single image capture as well as multiple image capture functionality and it is working great. Can you tell me the procedure to skip the image compression in iOS.

fubar:
@ivpusic Got a moment to look at this?

ivpusic:
thanks!

fubar:
Thanks for the merge! I have another PR in the pipes that adds an option to skip writing temporary files, just a heads up.

ios/ImageCropPicker.m
fubar: This will be available in the next release.

gianpaj:
Excuse my ignorance with ObjectC but what's the prop that can be set in JS component to disable image compression?

superandrew213:
@gianpaj just don't set the compress fields.

ryansolid:
@ivpusic Where did this land?

Is that still true? As far as I can tell if you don't set the compress fields doesn't it default to 0.8? Or the documentation atleast suggests as much. Is it possible to tell iOS not to compress the image file?

PR #564, opened 2018-01-07
Add support for all types of smart albums (selfies, time lapses, etc.)

ddim-honeybook:
Until now the user could only choose to show smart albums from the default set of 5 albums ('UserLibrary', 'PhotoStream', 'Panoramas', 'Bursts', 'Videos').

I fixed it and added support for all types of smartAlbums, as listed here:
https://developer.apple.com/documentation/photos/phassetcollectionsubtype/phassetcollectionsubtypesmartalbumgeneric

In the future we can add support for the other types of album subtypes, including the wildcard "any" (listed here: https://developer.apple.com/documentation/photos/phassetcollectionsubtype)

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ ImagePicker.clean().then(() => {
 | minFiles (ios only)                     |            number (default 1)            | Min number of files to select when using `multiple` option |
 | maxFiles (ios only)                     |            number (default 5)            | Max number of files to select when using `multiple` option |
 | waitAnimationEnd (ios only)             |           bool (default true)            | Promise will resolve/reject once ViewController `completion` block is called |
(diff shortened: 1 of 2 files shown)


ivpusic:
thanks! could you also update readme file with new types?

ddim-honeybook:
Sure! done.
The layout of the readme table changed a bit because I had to list down all of the possible values, see if you have an idea how to make it look nicer.

ivpusic:
thanks!

jasongaare:
Probably can close #552 based on this... 🙃

gianpaj:
Thanks for this!
Where could I find documentation on the `RecentlyAdded` option? How old can the images be?
Is it possible to restrict, for example, the user to pick only an image that was taken in the last 24-hours?

fubar:
Good stuff! FYI, you have a typo in the config value "Timepalses".

fubar:
Also just noticed (using the simulator) that `Animated` has no effect, but when I include `LongExposure`, both the "Long Exposure" and "Animated" albums show up.

PR #573, opened 2018-01-11
fix typo in creationDate property in response

jasongaare:
There was an extra colon in the response property for creationDate. There is no more ✅


ivpusic:
thanks!

PR #581, opened 2018-01-17
Added the additional gallery check to see whether its already open or not

svbutko:
It was possible to open the gallery twice in a row on Android when user would tap on the button which opens the gallery several times too fast. First I tried to implement it by simply making a Boolean variable but it didn't work as it should. And the only solution which I could come up with was to use the timer.

Same thing is not happening on iOS.


ivpusic:
does not look great to me, because we are relying on time component again. I suggest to implement it via javascript if you cannot think of a solution which does not rely on time component

svbutko:
Ok, I'll try to find better solution to it.

paintedbicycle:
This is happening in iOS too. I had it 3 or 4 times yesterday.

Issue #370, opened 2017-07-10
Can you change the "Move and Scale" header label text on the cropper?


larryranches:
Is there a prop where you can customize the label text of "Move and Scale" to something like "Pinch and Zoom"

### Version
Tell us which versions you are using: 

- react-native-image-crop-picker v0.13.1
- react-native v0.45.0

### Platform
- iOS
- Android

ivpusic:
currently not. PR welcome.

Ouadie:
This issue shouldn't be closed

ivpusic:
this is a feature request, not an issue :).

Ouadie:
yep you can reopen the issue and add a 'Feature Request' label and I'll do the PR.

ivpusic:
just send a pull request and reference this issue. Thanks a lot!

alainib:
+1 for this feature. i did'nt find how to open a new pull request, when i click on the button "new pull request" i'm redirected  to "https://github.com/ivpusic/react-native-image-crop-picker/compare"

Ouadie:
@ivpusic sorry for my late response, this request is no longer in our scope and we've finally used another component that satisfies our need. I'll try to fix it once I have some free time.

uinz:
@Ouadie can you share what Component did you using ?

Ouadie:
I'm not working on that project anymore, but the team are planning to use _uploadcare_ for uploading and cropping photo. 
You can ask @loiclouvet for more information about how to use _uploadcare_ with react native

mafiusu:
any update on that feature request? Or can anyone tell me where the title is listed in the node module, I would it change there.

ivpusic:
implemented here https://github.com/ivpusic/react-native-image-crop-picker/pull/601

PR #601, opened 2018-02-01
custom label for iOS

David-Owen-Walker:
Changes the "Move and Scale" header to the `cropperToolbarTitle` prop.

Solves https://github.com/ivpusic/react-native-image-crop-picker/issues/370


ivpusic:
thanks!

PR #602, opened 2018-02-01
Add option for writing temporary file

fubar:
Disk i/o is slow. When I use the `includeBase64` option to retrieve file contents directly, I have no use for a temporary file written to disk.

This PR adds a `writeTempFile` option (default true) that I can use to skip writing of those temporary files.

iOS only.


fubar:
@ivpusic Here's another small PR. Feedback appreciated. Thanks!

ivpusic:
thanks

PR #620, opened 2018-02-21
Added option to set free style crop for android.

NikolaRavic:
As uCrop (on which this lib relies for Android) supports this feature I thought it would be very nice feature to have this.


ivpusic:
could you also update readme file?

NikolaRavic:
Yes of course. In a moment ;)

ivpusic:
thanks!

NikolaRavic:
Thanks! That was fast

NikolaRavic:
Maybe just bump the version in package.json and publish new npm package so I can use new option.

Issue #645, opened 2018-03-13
Error when trying to get image from Google Drive


theohdv:
### Version

- react-native-image-crop-picker v0.20.0
- react-native v0.54.1

### Platform

- Android 7.0
- Real device

### Expected behaviour

Should return file path after selection.

### Actual behaviour

With the file picker, when I try to get a image from my google drive, an exception is thrown: "Error: Cannot resolve asset path."

### Steps to reproduce

1. Open android image picker

2. Open the left menu, click on drive and select an image.

### Attachments

<img width="541" alt="screen shot 2018-03-13 at 19 09 09" src="https://user-images.githubusercontent.com/26060802/37361081-4e03c4a0-26f2-11e8-88b8-a9f8f8161fdf.png">

### EDIT (14-03-18):
I've investigated and found that path is always null for cloud file (Android). I'm actually working on a PR.

theohdv:
Create PR #650

theohdv:
Maintaining a [fork](git@github.com:theohdv/react-native-image-crop-picker.git) with this feature and iCloud support.

isaachinman:
@theohdv Is there any update on this, to your knowledge? Why hasn't this been merged?

I just hit this issue with my QA team. How has there only been a single person to discover/work on this bug, for a lib with over 2k stars...? Seems very strange.

theohdv:
Fix with #650

PR #650, opened 2018-03-15
Add cloud support (Android)

theohdv:
**Current Behavior:**
If the user select an image in a cloud folder (drive for example) the picker return a path equal to null and throw an exception (see #645).

**Behavior expected with this PR:**
When the URI is local, the file is dowloaded and added to the cache directory. Nothing has changed. The only thing is that now an image from a cloud storage can be selected and is available in the local storage (cache).
The files downloaded are stored in the same cache folder, therefore `clean()`method will remove them.

**Tests:**
- Real device:
  Android 7.0

- Emulator:
  Android 8.0
  Android 4.4

### DIFF
--- a/android/src/main/java/com/reactnative/ivpusic/imagepicker/PickerModule.java
+++ b/android/src/main/java/com/reactnative/ivpusic/imagepicker/PickerModule.java
@@ -16,6 +16,7 @@ import android.provider.MediaStore;
 import android.support.v4.app.ActivityCompat;
 import android.support.v4.content.FileProvider;
 import android.util.Base64;
+import android.util.Log;
 import android.webkit.MimeTypeMap;
 
 import com.facebook.react.bridge.ActivityEventListener;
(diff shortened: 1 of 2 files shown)


superandrew213:
@theohdv did you test with both images and videos?

theohdv:
Yes I tried to get a video from my drive and it worked well.

superandrew213:
https://github.com/theohdv/react-native-image-crop-picker/blob/3ecb8388e84974345ee6100fd7fc6c9c1ba59249/android/src/main/java/com/reactnative/ivpusic/imagepicker/PickerModule.java#L327-L329
This needs to stay like this to support both videos and images.

Have you tested with dropbox or any other storage provider other than google drive?

theohdv:
I noticed that i could not just call `setType()` twice so I fixed it in the last commit.

About other storage I tried with google drive and one drive and it looks to work well.

Do you think this PR could be merge quickly ?

MyGuySi:
Do you have an ETA on this merge? The lack of support for cloud storage is holding me up in my current project

theohdv:
Not really, I'm currently using my fork. Still waiting for @ivpusic response for this PR.

ivpusic:
thanks a lot!

superandrew213:
I can confirm that dropbox fails with `column '_data' does not exist` error.

PR #655, opened 2018-03-22
Add UIDocumentPickerViewController -- iCloud support (iOS)

theohdv:
This PR introduce the `openFilePicker()` which should works the same way as the `openPicker()` (multiple selection not implemented yet). This method will let the user choose some image or video in the his iCloud storage using the iOS `UIDocumentPickerViewController`.

I tried to use the maximum amount of existing code that I could and to respect the general architecture of the project. I'm not a iOS developer and my knowledge in objective c are really low, so there is maybe some improvements that could be done.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,19 @@ ImagePicker.openCamera({
 });
 ```
 
+### Select from file (iOS only)
+
+This is not available on android as the android picker let the user select file from cloud storage.
+
+```javascript
+ImagePicker.openFile({
+}).then(image => {
+  console.log(image);
+});
+```
+
+**The `multiple` property has not been implemented yet -- coming soon**
+
 ### Crop picture
 
 ```javascript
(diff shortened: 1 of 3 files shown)


superandrew213:
There is `react-native-document-picker`. You can check that for the code. However you can just use that package to open the document picker. No need to add it to `react-native-image-crop-picker`.

What do you think @ivpusic?

theohdv:
I knew about `react-native-document-picker`, but it is not a good idea to me to add a dependency just  to support this feature (in addition only for iOS). This will make this project harder to maintain.

`react-native-image-crop-picker` already have cloud support for android (which was not working but has been fixed in my previous PR). So to add this feature would just put the iOS part at the same level.

I already have a look at `react-native-document-picker` code, but they do not use `dismissViewControllerAnimated ` so it didn't help me.

theohdv:
@ivpusic could you check this one plz ?

guy-shahine:
Any chance this PR will get reviewed and checked in?

theohdv:
Close as inactive

Issue #637, opened 2018-03-06
'RCTBridgeModule.h' file not found


Zidail:
### Version
- react-native-image-crop-picker v0.20.0
- react-native v0.51.0

### Platform

- iOS

### Expected behaviour

Links properly with no files missing.

### Actual behaviour

Gives a `Lexical or Preprocessor Issue` with `'RCTBridgeModule.h' file not found`.

### Steps to reproduce

1. Follow instructions in the current readme using Cocoapods.

Additional Notes: 
 * React is added via npm rather than pods.
 * Did not use `react-native link`
 * Already using XCWorkspace as I already have other pods in use.
 * Only added `pod "RNImageCropPicker", path: "../node_modules/react-native-image-crop-picker"
` to `Podfile`. (`:path => ` is equivalent to `path:`).

### Attachments

![Error](https://user-images.githubusercontent.com/9337042/37024517-0f568686-2164-11e8-8a15-9068fee564f2.png)

ivpusic:
you have to add

```
  rn_path = '../node_modules/react-native'

  pod 'yoga', path: "#{rn_path}/ReactCommon/yoga/yoga.podspec"
  pod 'React', path: rn_path, subspecs: [
    'Core',
    'RCTActionSheet',
    'RCTAnimation',
    'RCTGeolocation',
    'RCTImage',
    'RCTLinkingIOS',
    'RCTNetwork',
    'RCTSettings',
    'RCTText',
    'RCTVibration',
    'RCTWebSocket'
  ]
```

later you exclude it from the build anyway with

```
post_install do |installer|
  installer.pods_project.targets.each do |target|
    if target.name == "React"
      target.remove_from_project
    end
  end
end
```

Zidail:
AH! Thank you for answering me so quickly, @ivpusic 
Everything works now.

PR #656, opened 2018-03-23
Update RNImageCropPicker.podspec

chelongfei:
fix 'RCTBridgeModule.h' file not found
i found that in podsec, The lack of the  dependency about  'React/Core'
#637


justColbs:
This solved it for me. Please merge! Thanks.

nagad814:
Hi, React Pod is deprecated . Why are we using this method here. This is causing conflicts with other pod files

Issue #414, opened 2017-08-12
'RCTBridgeModule.h' not found


SeanLine:
### Version
Tell us which versions you are using: 

- react-native-image-crop-picker v0.15.1
- react-native v0.47.1

### Platform
Tell us to which platform this issue is related
- iOS
### Expected behaviour
Build succeeded

### Actual behaviour
'React/RCTBridgeModule.h'file not found
<img width="259" alt="2058577e-e50f-46d1-8950-76ecb4e1336c" src="https://user-images.githubusercontent.com/24886995/29241017-acd60ce6-7fa3-11e7-9d0e-b23ea50bd1bb.png">

i used cocoapods for installation of this library.

pureblood:
Doesn't seem to work out of the box.

Made it work by installing the dependencies via Cocoapods and Adding the imageCropPicker manually to the project i.e. add .xcodeproj file and link library

chuece:
@pureblood could you provide detailed instructions? 
how does your podfile look?
what do you mean under "link library"? adding imagecropPicker.a to "Linked libraries" under "general" doesn`t help at all

chuece:
ok, I finally got it working. Not sure this is a correct solution, but still, it works.
Step 1. yarn add https://github.com/ivpusic/react-native-image-crop-picker.git (this is for correct podspec if you need it, because it`s missing in the npm repository)
Step 2. drag imageCropPicker.xcodeproj from node_modules/react-native-image-crop-picker/ios into your workspace to "Libraries"
Step 3. add libimageCropPicker.a to "Linked frameworks and libraries" under "General" in your project
Step 4. open "Libraries/imageCropPicker.xcodeproj/ImageCropPicker.h" and change QBImagePickerController import to QBImagePicker import. 
Step 5. clean and build

pureblood:
@chuece I did the same exact thing. You explained it better.
I just added the QBImagePickerController via cocoapods. So I didn't have to alter any source file

chuece:
@pureblood but I`ve included QBImagePickerController in pods too, but there was an error "QBImagePickerController/QBImagePickerController.h" not found in ImageCropPicker.h. I believe pods is not set up correctly. But still I dont know how to make it work from pods

pureblood:
It should work if you have enabled `use_frameworks!` in your Podfile.

panhna:
i have the same problem on IOS too. android work well.

[...]

PR #657, opened 2018-03-23
Adds React as Dependency

kyledecot:
I was experiencing the issue that many people have been describing on https://github.com/ivpusic/react-native-image-crop-picker/issues/414. After looking at https://github.com/adjust/react_native_sdk/pull/22 it appears that we just need to add `React` as a dependency in the podspec.


kyledecot:
This has been fixed by https://github.com/ivpusic/react-native-image-crop-picker/pull/656 so I'm going to close this.

Issue #175, opened 2016-12-28
Orientation upload image


engin145:
### Version
- react-native-image-crop-picker v0.10.8
- react-native v0.36.1

### Platform
- iOS

When i upload images i take with orientation = 6

ivpusic:
can you please explain a little bit better what the problem is?

engin145:
Image head information has field orientation from 1 to 8. The image that I upload to my server is maintained with the orientation parameter 6 is not correctly displayed, and turned 90 degrees counterclockwise. We need to do so that would be the picture was with the orientation parameter 1 if it is flat.

ivpusic:
Image orientation is part of exif jpeg data. I had similar problem in my project, but the problem was not this library, but server code. If you are also using Go on backend to process images, then here is the problem. If you are not using Go, please verify that processing on server is not wrong => orientation data is lost.

engin145:
My server side takes a image and compresses it in the same orientation in which the load. And already loaded upside down. I do not use Go.

engin145:
I saw a similar decision in the library with correcting images orientation https://github.com/marcshilling/react-native-image-picker/blob/master/ios/ImagePickerManager.m#L603

ivpusic:
would you be interested to submit PR?

ivpusic:
I tried to reproduce this, but cannot. If you are still having issues with this, please upload image which is having problem.

gooddaddy:
I have the same problem in react-native-image-crop-picker v0.20.1.

I find out a place where you could make a mistake.

source code:
react-native-image-crop-picker/ios/src/ImageCropPicker.m

- (void) processSingleImagePick:(UIImage*)image withExif:(NSDictionary*) exif withViewController:(UIViewController*)viewController withSourceURL:(NSString*)sourceURL withLocalIdentifier:(NSString*)localIdentifier withFilename:(NSString*)filename withCreationDate:(NSDate*)creationDate withModificationDate:(NSDate*)modificationDate{
   .....
  ImageResult *imageResult = [self.compression compressImage:image withOptions:self.options];
   .....
}

I think the right way to do is：
   ImageResult *imageResult = [self.compression compressImage:[image fixOrientation] withOptions:self.options];

wonday:
@superandrew213 
maybe not related, But I use master to test #681,
It still has problem.

PR #682, opened 2018-04-21
Fix orientation

superandrew213:
Fixes https://github.com/ivpusic/react-native-image-crop-picker/issues/175#issuecomment-383274304


ivpusic:
tnx!

PR #684, opened 2018-04-23
Update TypeScript typings with some Android-only options

richeterre:
Several Android-only options were missing from the TypeScript `Options` interface, causing compiler warnings in client codebases when used.


ivpusic:
tnx!

Issue #689, opened 2018-05-03
Corrupt JPEG Written to Temp Storage on iOS 11.3 if `compressImageQuality` Unset


medelman17:
### Version
Tell us which versions you are using: 

- react-native-image-crop-picker v0.20.1
- react-native v0.54-55.2

### Platform
iOS 11.3

### Expected behaviour

RNICP should write universally valid JPEG files to temporary storage at path image.path when user selects HEIC image(s) from camera roll regardless of whether `compressImageQuality` is set. 

### Actual behaviour

When a HEIC image is selected on a real device in multi-picker, the temporary file written to image.path is partially corrupt unless compressImageQuality is set.  Images written to image.path without compressImageQuality setting are reported as invalid by my image-processing pipeline, which uses Sharp. Further, these images cannot be displayed in a browser using an image tag (although they can be downloaded and viewed in a standalone image viewer--preview in my case). This does not occur if the image is captured by RNICP's camera. Adding compressImageQuality to the ImagePicker.openPicker options cures the problem. 

I'm not sure when this bug was introduced or how, but I was updating a currently deployed app that uses RNICP without specifying the compressImageQuality setting and have not run into this issue previously. 

### Steps to reproduce

1. Select HEIC image from iPhone running iOS 11.3 using `ImagePicker.openPicker({ multiple: true, includeExif: true })`. 

2. Upload jpg file at image.path to S3 using a signed URL with XHR specifying image/jpeg as the content type. 

3. Attempt to view image in browser and/or run any sort of post-processing on the image. 

4. Pull hair out for ~6 hours. 

### Attachments
![screen shot 2018-05-03 at 7 01 51 pm](https://user-images.githubusercontent.com/14793389/39606718-cfb79b54-4f04-11e8-9a10-0addf5688a33.png)

// stacktrace or any other useful debug info

Error: Input buffer contains unsupported image format

Love react-native-image-crop-picker? Please consider supporting our collective:
👉  https://opencollective.com/react-native-image-crop-picker/donate

medelman17:
I will certainly give it the old college try; however, I’m no Native coder. That said, I’ve always wanted to learn more. So, will try and figure it out this weekend.

fubar:
[...]

Issue #681, opened 2018-04-21
Multiple Picker return error image


wonday:
### Version
Tell us which versions you are using: 

- react-native-image-crop-picker v0.20.1
- react-native v0.53.4

### Platform
Tell us to which platform this issue is related

- iOS

### Expected behaviour
Select multiple images from iPhone gallery with original size,
if jpeg image exif's orientation is not normal, the returned image file can not show.
But if use single picker, it will OK. 

in ImageCropPicker.m
```
- (void)qb_imagePickerController:(QBImagePickerController *)imagePickerController  didFinishPickingAssets:(NSArray *)assets {
...
                                 if (isLossless && useOriginalWidth && useOriginalHeight) {
                                     // Use original, unmodified image
                                     imageResult.data = imageData;
                                     imageResult.width = @(imgT.size.width);
                                     imageResult.height = @(imgT.size.height);
                                     imageResult.mime = [self determineMimeTypeFromImageData:imageData];
                                     imageResult.image = imgT;
//this block not OK                        
                                 } else {
//this block OK
                                     imageResult = [self.compression compressImage:[imgT fixOrientation] withOptions:self.options];
                                 }
.....
```

### Steps to reproduce

```
        ImageCropPicker.openPicker({
            mediaType: 'photo',
            multiple: true,
            waitAnimationEnd: false,
            maxFiles: 9
        }).then(images => {
            ...
        }
```
### Attachments
the attachment is an image from my iPhone that can not show.
[abc.zip](https://github.com/ivpusic/react-native-image-crop-picker/files/1934224/abc.zip)

fubar:
Hi @wonday, can you provide more context? I don't know what you mean by "can not show". This is an image picker library, not one that displays images. I cannot reproduce this issue and image you attached looks fine. Is it possible that there is an issue with how you are displaying images after the work of the image picker is complete?

PR #735, opened 2018-06-26
Convert unknown file types to jpeg

fubar:
@ivpusic As discussed in https://github.com/ivpusic/react-native-image-crop-picker/issues/689, this change will convert unknown image formats to JPEG.

Possibly also addresses https://github.com/ivpusic/react-native-image-crop-picker/issues/681


ivpusic:
thanks!

VLNTNA:
@fubar for me it seems that on iOS this feature only works when multiple image selection is enabled as the relevant code within `qb_imagePickerController` is only executed when `multiple` is `true`.

am I missing something? is there a way to make this work for single image picking?

fubar:
@VLNTNA you're right. There's an architectural issue here where some business logic is duplicated between the "multiple" or "single" branches which have now diverged. This is a more overarching problem that has been around for a while. Maybe @ivpusic can give more insight.

PR #736, opened 2018-06-26
Reduce default compressed file size

fubar:
@ivpusic This is a suggestion. As previously discussed (don't specifically remember where), the sweet spot between quality and file size is somewhere in the 70-80% range. In my tests it was hard to see a noticeable increase in image quality beyond 80%, while at 80% the size of the output file is about half or less compared to a quality setting of 100%. As a result I believe 80% is a decent default.

I can only test this on iOS and I'm not sure if the same behaviour exists on Android too.

Related issues: https://github.com/ivpusic/react-native-image-crop-picker/issues/449, https://github.com/ivpusic/react-native-image-crop-picker/issues/252

I've also added a hint to the readme for how to use base64 image data to display images as this seems to be a repeat question and should make it easier for folks.


ivpusic:
thanks!

Issue #356, opened 2017-06-23
Scale image bigger than the crop window?


binoy14:
### Version
Tell us which versions you are using: 

- react-native-image-crop-picker v0.12.9
- react-native v0.44.0

### Platform
Tell us to which platform this issue is related

- iOS

First of all, thank you for the great library! I am wondering if there is a way to scale out an image if it's bigger than a crop window. For instance if I set the crop window to 375x375 and if the image picked is 1080x375 is there a way to zoom out so the full image can be shown?

AunMSheikh:
If you are looking to get your hands dirty in the node_module folder there is a  folder react-native-image-crop-picker you will find the file ImageCropPicker.m in the ios folder. In that file is where all the views for cropping are created you can try changing the dimensions there and see if you get the desired results. I suspect the method imageCropViewControllerCustomMaskRect is probably where you'd want to change the dimensions. Good luck!

ivpusic:
yes. will close this one as for now this is out of scope for the library.

PR #754, opened 2018-07-05
Added `cropperAvoidEmptySpaceAroundImage` param

diegolmello:
This PR makes possible to choose a full image (without cropping) using `cropping: true`.

It also fixes a bug causing unnecessary resizes after a crop.
This would create poor image quality.

Related to #356


diegolmello:
Another PR accepted with these changes

keshamtravels:
So , does this correct the issue of very poor quality of image after crop , should i make updates in my app for this repo

diegolmello:
@keshamtravels Yeah, I messed up this PR by mixing subjects.
It does two things: adds `cropperAvoidEmptySpaceAroundImage` and fixes that poor quality image.
If @ivpusic is ok with it, I can open another PR just to fix the second one.

ivpusic:
@diegolmello sounds good. could you create a separate PR which fixes image quality issue after cropping? thanks

diegolmello:
@ivpusic Sure! I'll take a look.

PR #768, opened 2018-07-17
Add new logo/icon

mansya:


ivpusic:
tnx a lot! could you make image which is on readme file as an SVG?

ivpusic:
thanks a lot!

Issue #134, opened 2016-11-10
Need localIdentifier for Video assets when using compressVideo = false


kviswanath:
### Version
Tell us which versions you are using: 

- react-native-image-crop-picker v0.8.3
- react-native v0.32

### Platform
Tell us to which platform this issue is related

- iOS

### Expected behaviour

Package should return a unique identifier for the asset which can be used to upload the asset to a remote web server. Using classes like PHAsset requires localIdentifier of the asset.

Any other means of querying the selected asset based on path will be of great help.

### Actual behaviour

The data returned by the package does not include the "localIdentifier" or some other unique identifier for the asset.

There is no way to read and upload the selected asset directly. The only option is to set compressVideo to true (default value) and allow copying the asset to the tmp directory. This becomes accessible to the app for upload. However if someone wants to upload the original video without any compression, that is not possible with this package.

### Steps to reproduce

1. Set compressVideo to false. This does not copy the video asset to a tmp directory.

2. Select a Video from the picker.

3. Try to upload the asset using the path returned by the package.

4. Will get a "read permission error" because iOS does not allow accessing Photo assets using NSData.

kviswanath:
@ivpusic Any comments on the requirement?

ivpusic:
@kviswanath sorry for the delay. However currently I am a little bit in rush. Do you have a time to send PR? Attaching useable URL should not be big task.

kviswanath:
@ivpusic sorry for the delay. I am thinking of the doing the following changes, please let me know your thoughts on the same.
* Add an additional attribute "asset_identifier" to the output
* Read the attribute localIdentifier from the PHAsset object.
* Change the signature of the method createAttachmentResponse and add another attribute "withAssetIdentifier".
* Pass the asset_identifier as nil when compressVideo attribute is false.
* Change the Readme file to include this iOS only output attribute.

ivpusic:
hm, if we can upload video by using "localIdentifier" value, can we just change https://github.com/ivpusic/react-native-image-crop-picker/blob/69d3ed45048a70f8cdbd84c6f490031f4f68f355/ios/ImageCropPicker.m#L296 to include "localIdentifier" value instead of current `[sourceURL absoluteString]` which does not work when uploading. This way users will just get single url, which works, and they won't have to know about which url when to use.

kviswanath:
Giving the localIdentifier (this is just a unique id and does not look like a path) will enable the user to upload the original image/ video.

[...]

PR #769, opened 2018-07-18
Add boolean `compressVideo` option (iOS only)

tiagoalves:
Following up to #134, I need exactly the same functionality as described in [this comment](https://github.com/ivpusic/react-native-image-crop-picker/issues/134#issuecomment-268145969). That is, I need to perform the video compression step as part of another module with a custom progress UI.

The best option until now was to use `compressVideoPreset: "Passthrough"` but that is still slow if you select multiple larger videos (can take several minutes) and consumes unnecessary disk space, even if temporary.

This new `compressVideo` boolean option, when set to false, allows us bypass the video compression step. We can then work with the selected files very easily with the `localIdentifier` property that is returned.

Note that with `compressVideo: false`, the returned `path` property is `null`. Not sure if that makes sense.

If you need any change or improvement, just let me know.


ivpusic:
thanks a lot for your PR. Could you provide example of `We can then work with the selected files very easily with the localIdentifier property that is returned.`, probably something like this can be useful to other people looking for a similar functionality

PR #771, opened 2018-07-19
Use project-wide properties and new dependency

KarlosQ:
* Support for gradle 3.+

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -1,12 +1,17 @@
 apply plugin: 'com.android.library'
 
-android {
-    compileSdkVersion 27
-    buildToolsVersion "27.0.3"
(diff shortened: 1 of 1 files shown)


ivpusic:
thanks

superandrew213:
@ivpusic this breaks builds for projects using gradle 2.+ since `implementation` is not recognised:

```
implementation 'com.facebook.react:react-native:+'
implementation 'com.github.yalantis:ucrop:2.2.2-native'
implementation 'id.zelory:compressor:2.1.0'
```

React native is still using gradle 2 by default and gradle 3 still supports `compile`. Can we revert this part of this PR?

ivpusic:
sure, do you have some minutes to send a PR?

Issue #642, opened 2018-03-09
iPhone hangs on processing assets when uploading vidoes


L-Yeiser:
### Version
Tell us which versions you are using: 

- react-native-image-crop-picker ^0.20.0
- react-native 0.53.0

### Platform
Tell us to which platform this issue is related

- iOS
- Android

### Actual behaviour
Uploading a video on an iphoneX hangs on processing assets. Works on the IOS simulator and on android devices.

### Steps to reproduce

1. Open image picker.

2. Select a video from the camera roll.

3. Process hangs.

No errors are logged.

code: 
```
 selectImage() {
    ImagePicker.openPicker({
      multiple: true,
      mediaType: 'any',
    }).then((media: ImagePickerUpload[]) => {
      console.log(media);
      this.setState({
        ...this.state,
        media,
      });
    });
  }
```

L-Yeiser:
As a follow up. I do not have this issue with an iphone 6 nor an iphone 7.

ivpusic:
do you have some time to investigate what is happening here?

L-Yeiser:
I am under a pretty strict deadline at the moment and am pretty new to react-native so do not have the time or resources to really dig in. If I find time this weeks I will see what I can come up with.

L-Yeiser:
Our team has sunk a bit of time into this and come up empty handed. Other then knowing it is an IOS 11+ issue. We thought it had to do with an `invalid response for blob type` error but that was fixed with `RN 0.55` but that is not the case. For now we are using `react-native-image-picker` when a user is running IOS 11 + but it would be awesome to get this library working with newer IOS.

L-Yeiser:
This issue was resolved when we upgraded to
`react-native 0.55.0`
`react-native--image-crop-picker 0.20.3`

chrisbobbe:
I can confirm that I **only** have this issue when I set `multiple: true` **and** I try to select multiple videos.

liuyunzyj:
Has anyone come across this problem?
![image](https://user-images.githubusercontent.com/40293335/61839310-afa01000-aebf-11e9-8041-05c5960b4e9b.png)
![image](https://user-images.githubusercontent.com/40293335/61839337-c21a4980-aebf-11e9-9fd3-eb8b643bad86.png)

Bashamd455:
Hi @chrisbobbe, Even I am facing the same issue i.e., iPhoneX taking forever to process a video when selecting from the camera roll.
[...]

PR #772, opened 2018-07-20
Solves #642 by allowing network access for videos stored in the cloud

chrisbobbe:
Set options.networkAccessAllowed = YES in getVideoAsset. This will prevent freezing on 'Processing Assets...' when attempting to get videos stored in the cloud.

I referred to https://stackoverflow.com/questions/40446026/phcacheimagemanager-requestavassetforvideo-returns-nil-avasset?rq=1 when investigating this bug. I also noticed that options.networkAccessAllowed was set to YES for photos. So I just set it to YES for videos too.

Thanks for this awesome project!


ciriac:
I tested this and it fixes the issue that we're having. It's been a month now, any reason why this hasn't been merged in?

PR #791, opened 2018-08-09
Adds forceJpg flag to convert live photos

migue1s:
Adds a `forceJPG` flag to force the lib to convert target images. This will be mostly helpful when images contain Live Images but only Still images are wanted.

How it works:
Calling `UIImageJPEGRepresentation` will force any images to return their JPG equivalent, even if they're Live Photos.

### DIFF
--- a/example/app.js
+++ b/example/app.js
@@ -134,7 +134,7 @@ export default class App extends Component {
       multiple: true,
       waitAnimationEnd: false,
       includeExif: true,
-      forgeJPG: true,
+      forgeJpg: true,
     }).then(images => {
       this.setState({
         image: null,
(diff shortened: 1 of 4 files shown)


rajputav:
nicee 💯

SonnyBrooks:
💯

ivpusic:
thanks a lot! could you rename `forceJPG` to `forceJpg`?

ivpusic:
thanks a lot!

example/app.js
```diff
@@ -134,6 +134,7 @@ export default class App extends Component {
       multiple: true,
       waitAnimationEnd: false,
       includeExif: true,
+      forgeJpg: true,
```
fubar: You have a typo here in "forge"

migue1s: Great catch, will do a follow-up PR

Issue #795, opened 2018-08-21
A problem occurred evaluating project ':react-native-image-crop-picker'. > Could not find method implementation() for arguments [com.facebook.react:react-native:+] on object of type org.gradle.api.internal.artifacts.dsl.dependencies.DefaultDependencyHandler.


vkandk:
### Version
Tell us which versions you are using: 

- react-native-image-crop-picker v0.21.0
- react-native v0.55.2

### Platform
Android

### Expected behaviour
Trying to run on android device with the command. 
npm run android

### Actual behaviour
> node node_modules/react-native/local-cli/cli.js run-android

Scanning folders for symlinks in ..\node_modules (172ms)
Starting JS server...
Building and installing the app on the device (cd android && gradlew.bat installDebug)...
Incremental java compilation is an incubating feature.

FAILURE: Build failed with an exception.

* Where:
Build file '..\node_modules\react-native-image-crop-picker\android\build.gradle' line: 23

* What went wrong:
A problem occurred evaluating project ':react-native-image-crop-picker'.
> Could not find method implementation() for arguments [com.facebook.react:react-native:+] on object of type org.gradle.api.internal.artifacts.dsl.dependencies.DefaultDependencyHandler.

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.

BUILD FAILED

### Steps to reproduce

1. build.gradle

// Top-level build file where you can add configuration options common to all sub-projects/modules.

buildscript {
    repositories {
        jcenter()
    }
    dependencies {
        classpath 'com.android.tools.build:gradle:2.2.3'

        // NOTE: Do not place your application dependencies here; they belong
        // in the individual module build.gradle files
    }
}

allprojects {
    repositories {
        mavenLocal()
        jcenter()
        maven {
            // All of React Native (JS, Obj-C sources, Android binaries) is installed from npm
            url "$rootDir/../node_modules/react-native/android"
        }
        maven { url "https://maven.google.com" }
        maven { url "https://jitpack.io" }
    }
}

2. android/app/build.gradle

apply plugin: "com.android.application"

import com.android.build.OutputFile

/**
 * The react.gradle file registers a task for each build variant (e.g. bundleDebugJsAndAssets
 * and bundleReleaseJsAndAssets).
 * These basically call `react-native bundle` with the correct arguments during the Android build
 * cycle. By default, bundleDebugJsAndAssets is skipped, as in debug/dev mode we prefer to load the
[...]

PR #796, opened 2018-08-21
Revert to compile

superandrew213:
Revert back to using `compile` to continue support for gradle 2.
 
Fixes https://github.com/ivpusic/react-native-image-crop-picker/issues/795


armanatz:
@ivpusic This needs to be temporary as `compile` will be removed by the end of 2018 and it will break builds after that.

PR #809, opened 2018-08-30
Fix typo on example project for forceJpg

migue1s:
The example project had a small typo, this is fixed by this PR. I'm currently using the actual implementation of the feature and it works as expected (apart from the example).

This was signaled in a comment in #791


ivpusic:
thanks

Issue #670, opened 2018-04-08
Error: column '_data' does not exist [Android] [BUG]


isaachinman:
### Version

- react-native-image-crop-picker v0.20.1
- react-native v0.54.2

### Platform

- Android

### Expected behaviour

Picking from third party apps like Dropbox works

### Actual behaviour

`ImagePicker.openPicker` throws an error on selection:

    E/image-crop-picker: Promise rejected. column '_data' does not exist

### Steps to reproduce

1. Open picker:

```
    const images = await ImagePicker.openPicker({
      multiple: true,
      includeExif: true,
      maxFiles: 15,
      mediaType: 'photo',
      smartAlbums: ['UserLibrary', 'RecentlyAdded', 'PhotoStream', 'Generic', 'Favorites', 'SelfPortraits', 'LongExposure'],
    })
```

2. Go to Dropbox

3. Select one or more photos

4. Click "Done"

### Attachments

![screen shot 2018-04-08 at 17 56 55](https://user-images.githubusercontent.com/10575782/38469514-47435f04-3b56-11e8-8aea-eee81d790a8a.png)

### Notes

I am using gradle `3.1.0-alpha09`. Not sure if this has something to do with the issue, but picking from my regular camera roll works just fine on Android.

superandrew213:
@isaachinman did you find a fix or workaround for this? Getting this issue for Dropbox and Google Drive.

isaachinman:
@superandrew213 I abandoned this issue after reporting it due to time constraints.

@ivpusic Do you mind explaining why you just closed the issue?

isaachinman:
Ah sorry guys, just followed the PR link through. Very fast work! Thanks @superandrew213.

PR #817, opened 2018-09-07
Fall back to writing to file if _data column does not exist

superandrew213:
Fixes https://github.com/ivpusic/react-native-image-crop-picker/issues/670


ivpusic:
thanks!

Issue #61, opened 2016-09-05
Incomplete bit code when uploading to app store


superandrew213:
### Version
Tell us which versions you are using: 

- react-native-image-crop-picker v0.8.0
- react-native v0.32.0

### Steps to reproduce

1. Archive and upload to app store

### Attachments
<img width="870" alt="screen shot 2016-09-05 at 6 51 09 pm" src="https://cloud.githubusercontent.com/assets/13059204/18242518/18f4cef6-739a-11e6-95ce-948b695e143d.png">

superandrew213:
Getting this now after cleaning, restarting, reinstalling, etc

<img width="877" alt="screen shot 2016-09-05 at 7 05 05 pm" src="https://cloud.githubusercontent.com/assets/13059204/18242944/097cd02a-739c-11e6-9c1a-c31bdd24ff4d.png">

ivpusic:
you are getting this error because pre-built frameworks are built for simulator and device ARCHs. This is good for development because you don't have to worry about supported arch when building and running on different platforms. 

Before uploading to appstore, you have to strip away simulator platforms, and then use stripped versions of frameworks instead of current ones.

You can also remove current pre-built libraries for embedded binaries, and add the ones from `Libraries/imageCropPicker/Libraries/XXX/Product/XXX.framework` which will be there after you build project for specific platform.

Mentioum:
@ivpusic could you give a step by step for this process as I'm having real trouble getting rid of this error.  It's definitely due to a misunderstanding of mine of how xcode works.

ivpusic:
then probably latter will be easier for you. 

go to project General tab, scroll down to Embedded binaries. You will see that there are two pre-built frameworks. You can remove them, build project for Device => this will produce Frameworks built for Device platform. Click + under Embedded binaries, and now you can add these two frameworks.

<img width="344" alt="screen shot 2016-09-05 at 6 15 54 pm" src="https://cloud.githubusercontent.com/assets/450140/18253439/f2e94b10-7394-11e6-9c56-c9a61642976a.png">

Later probably you will want to switch back to pre-built frameworks because it is easier during development.

superandrew213:
@ivpusic I'm still getting the same error if I do that.

superandrew213:
This is what I did:

1. Completely remove react-native-image-crop-picker
2. Install and link react-native-image-crop-picker
3. Build for "Generic iOS Device"
[...]

PR #821, opened 2018-09-11
Remove problematic step from manual post-install steps for iOS

mdio:
I verified that this step ("drop the ios/ImageCropPickerSDK folder to your xcode project") is not required using Xcode 9.4.1, iOS 11.4 and React Native 0.55.4 when doing a manual install.

Adding the ImageCropPickerSDK folder would also trigger the "CFBundleIdentifier Collision" error on TestFlight as reported in issue #61.

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,7 +245,6 @@ In Xcode open Info.plist and add string key `NSPhotoLibraryUsageDescription` wit
 
 ##### Only if you are not using Cocoapods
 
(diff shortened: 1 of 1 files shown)


origamih:
Confirmed work for Xcode 10, React Native 0.55. You don’t need Step 2 but you need step 3 (Post install)

The wonderful thing is you don’t have to do anything for AppStore publishing

ivpusic:
tnx!

PR #825, opened 2018-09-14
[FIX] Poor image quality after cropping

diegolmello:
@ivpusic We talked about this at #754.

On iOS, we were cropping the image twice.
On Android, image size were default 200x200.

Thanks!

### DIFF
--- a/android/src/main/java/com/reactnative/ivpusic/imagepicker/PickerModule.java
+++ b/android/src/main/java/com/reactnative/ivpusic/imagepicker/PickerModule.java
@@ -87,8 +87,8 @@ class PickerModule extends ReactContextBaseJavaModule implements ActivityEventLi
 
     //Light Blue 500
     private final String DEFAULT_WIDGET_COLOR = "#03A9F4";
-    private int width = 200;
-    private int height = 200;
+    private int width = 0;
+    private int height = 0;
 
     private Uri mCameraCaptureURI;
     private String mCurrentPhotoPath;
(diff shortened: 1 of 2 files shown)


keshamtravels:
Please merge this pull request asap and what i would need to do if i want to get latest changes in my app for this repo

evaleirasclapps:
Merge please! I need this

ivpusic:
@diegolmello merged and released https://github.com/ivpusic/react-native-image-crop-picker/releases/tag/v0.21.2

diegolmello:
@ivpusic Why haven't you accepted my PR? As this is a community, it would be nice to get my name as contributor (not just on the release).

ivpusic:
I appologize for that @diegolmello, I did some small changed based on your fork in a seaparate branch, and pushed everything together to master

evaleirasclapps:
When is going to be released to NPM??

ivpusic:
it is already released in version 0.21.2

evaleirasclapps:
This did not work for me, I added `compressImageQuality: 1` and it's still compress the image TOO MUCH

haseebnaseem:
The fix does does not work.

haseebnaseem:
Did anyone solve this ? The fix works on iOS but not android

SkariDamm:
I was facing the same problem on Android when using the device dimensions as crop dimensions options for `ImagePicker.openCamera`.

I was successful in retaining the full image quality by doing:

```
const image = await ImagePicker.openCamera({});
const croppedImage = await ImagePicker.openCropper({
    path: image.path,
    width: image.width,
    height: (image.width * 9) / 16
});
```

diegolmello:
This PR works in my case, because I don't pass width and height props.
It was resizing, cropping, compressing and resizing again.
It may cause issues if you pass width and height, as pointed out.

Lowisoft:
Yes, but it's a pity that we can't use fixed width and height props after this PR (which is an essential part of cropping) and that there might be other people who will run into this issue too.

Lowisoft:
@diegolmello Unfortunately, I am not very familiar with Swift. 
Does the changed code (for iOS) only fix the double cropping? In fact, without these changes everything works as expected and I didn't notice that the double cropping or quality caused any problems on a real iOS device.

PR #826, opened 2018-09-14
Update README.md to add ios configuration for default Camera/Gallery button localization

ishigamii:
Add optional configuration details about default ios text button for the camera and the gallery


ivpusic:
thanks

PR #855, opened 2018-10-21
Russian localization for iOS

pinguinjkeke:


ivpusic:
it is better to submit this PR to QBImagePicker repository, because the one in this repo is just a copy which can be replaced anytime.

pernifin:
@ivpusic Why did you approve French locale PR ([https://github.com/ivpusic/react-native-image-crop-picker/pull/593](url)) but closed this one? 
As QBImagePicker has not been updated for years, I don't see any reason to keep local version clean.
I actually have to switch to another RN image picker now, simply because need Russian locale.

PR #862, opened 2018-10-25
Update PickerModule.java

bartoszwolski:
Fix for default values being kept when calling cropPicker from multiple locations.

Now default values if not passed will be reverted to default.


ivpusic:
released here https://github.com/ivpusic/react-native-image-crop-picker/releases/tag/v0.21.3

Issue #872, opened 2018-11-06
[Android/iOS] Different result for OpenCamera


ishigamii:
### Version
Tell us which versions you are using: 

```
"react-native-image-crop-picker": "^0.21.1",
"react-native": "0.55.4",
```

### Platform
- iOS
- Android

### Expected behaviour
The output given in the success promise of `openCamera` should be the same between Android and iOS.  

### Actual behaviour
Actually the output of image on Android is an `Array of one element` whereas on iOS we have directly the `Object of the image`. 

```
 ImagePicker.openCamera({
      multiple: true,
      maxFiles: NB_IMAGE_MAX - this.state.medias.length,
      mediaType: 'photo',
    })
      .then((image) => {
        console.log('Image', image);
        this.showPicture(image);
      })
      .catch((cancel) => {
        console.log('Cancel Camera', cancel);
      });
```

Would be nice to have the same output for iOS and Android
Thanks a lot

PR #873, opened 2018-11-06
Change resolve return type for openCamera to Array as in Android

ishigamii:
Should correct what I specified in this issue : https://github.com/ivpusic/react-native-image-crop-picker/issues/872

Don't know why it also contains my old PR with the readme update that is already up to date (sorry about that)


ishigamii:
@GuillotJessica not sure it is a feature they want but seems pretty fair

ivpusic:
changed in https://github.com/ivpusic/react-native-image-crop-picker/releases/tag/v0.22.0. now both ios and android return object when calling openCamera

PR #891, opened 2018-11-29
Added missing flags to support front camera on some android phones

juanjonucleus:
I did some searching and there are different flags for different phones, in my case previous flag for Android was not working with a Google Pixel phone, Im adding 2 new flags tu support more android phones


ivpusic:
thanks!

PR #918, opened 2019-01-19
Simplify `includeBase64`

mieszko4:
If `image.data` is already base64 encoded string, why do we need this wrapper?
The proposed change worked for me on ios and android.


ivpusic:
tnx!

PR #949, opened 2019-03-01
add pickByAlbum params

artzok:


ivpusic:
tnx for PR. can you explain what this option is doing?

PR #962, opened 2019-03-17
Update to AndroidX

rafcontreras:


ovy9086:
why don't you squash this commits ? what's the point of having 3 commits that are later just being reverted 🤷‍♂️

RichardLindhout:
@ovy9086 What is your point? He is probably using the inline edit tool from Github so you don't have to manually add a branch etc.

Saad-Bashar:
When can we have the support for androidx?

Saad-Bashar:
+1

se-bastiaan:
These changes are nice, but for the package to be completely AndroidX compatible the ucrop dependency should be updated: https://github.com/Yalantis/uCrop/pull/538

ivpusic:
released in 0.25.0

Issue #379, opened 2017-07-18
[Android] When selecting image from gallery or taking photo, image loses orientation


joeferraro:
### Version
Tell us which versions you are using: 

- react-native-image-crop-picker v0.14.1
- react-native v0.43.1

### Platform

- Android

### Expected behaviour

- Take photo or select photo from gallery
- Image is saved locally **with** original orientation

### Actual behaviour

- Take photo or select photo from gallery
- Image is saved locally **without** the original orientation

### Steps to reproduce

```
ImagePicker.openPicker({
  cropping: false,
  includeBase64: true,
  compressImageQuality: 0.5,
});
```

**Important** This only occurs when `cropping` is set to `false` and `compressImageQuality` is set to a value less than `1`.

ivpusic:
do you have some time to take a look into this?

joeferraro:
@ivpusic this issue seems related to [Compressor](https://github.com/zetbaitsu/Compressor). [`v2.1.0`](https://github.com/zetbaitsu/Compressor/commit/d69833452767c239f6f5b29a2dc0aca080cf11b2) of Compressor attempts to resolve the issue but there seems to be a regression for compressed images taken by the camera [as I've noted here](https://github.com/zetbaitsu/Compressor/issues/64#issuecomment-316181036).

ferminmoli:
Is there any news about this? Image is loosing it's orientation

ferminmoli:
I updated to 0.16.0 version (The last one) and Taking photo from camera is crashing on Android. Please, I need help with this.

WLYau:
Hi, I'm experiencing the same issue, have updated to the latest version and the image still gets rotated? Anymore news on this?

We're using react-native version 0.39.2

dinatazy:
Happens also with me on IOS. Image is rotated upside down. Any updates on this issue?

Clowning:
Any updates ?

dinatazy:
@Clowning   It turns out it's not the library's issue.  my solution was that I get the Exif data by setting property includeExif to true.  the exif object includes an orientation number. Each number indicates a specific orientation so , based on the orientation number you can rotate the image to its original position again.

isaachinman:
@dinatazy I just hit this issue myself. Strange that more people haven't encountered it. When an image has this `Orientation` EXIF field set, `react-native-image-crop-picker` will actually swap the image dimensions, but leaves the rotation itself up to the end user.

[...]

PR #966, opened 2019-03-20
add option to respect exif rotation data when selecting an image on Android

JPig:


karltaylor:
This actually fixed my rotation issue as mentioned here https://github.com/ivpusic/react-native-image-crop-picker/issues/379#issuecomment-516343409

eubnara:
If `compressImageQuality` is not 1.0, it makes rotate twice.

https://github.com/ivpusic/react-native-image-crop-picker/blob/master/android/src/main/java/com/reactnative/ivpusic/imagepicker/Compression.java#L40-L42

atfilipcc:
Patching this into 0.38.0 solved my issues, thank you 🎉

PR #973, opened 2019-03-28
Add some types for Image interface

stonezzang:
In typescript project, we can not access directly some properties.
`filename`, `creationDate`, and `modificationDate` are useful properties for user.


ivpusic:
thanks!

Issue #905, opened 2018-12-20
WRITE_EXTERNAL_STORAGE needed for SDK >= 19?


mattfysh:
### Version
Tell us which versions you are using: 

- react-native-image-crop-picker v0.21.3
- react-native v0.57.4

### Platform
Tell us to which platform this issue is related
Android

### Question

From SDK version >= 19, `WRITE_EXTERNAL_STORAGE` permission is no longer needed, if you're reading and writing to/from an app-specific location, from the docs:

> Starting in API level 19, this permission is not required to read/write files in your application-specific directories returned by  `getExternalFilesDir(String)` and `getExternalCacheDir()`.

Link: https://developer.android.com/reference/android/Manifest.permission?hl=es#WRITE_EXTERNAL_STORAGE

Given that this library is only writing to its own location, and not needing read or write access to the remaining filesystem, this permission could be safely removed (or replaced with `READ_EXTERNAL_STORAGE` for accessing the photo gallery)

RHackrid:
Same problem in following use-case: select multiple photos (without cropping/editing)
`ImagePicker.openPicker({
        multiple: true,
        mediaType: "photo"
    });`

The App does not need the permission `WRITE_EXTERNAL_STORAGE` but without this permission the previous call will return following error: Required permission missing

Please check the write-permissions only in case of writing.
Link: [https://developer.android.com/training/permissions/usage-notes](url)

The permission `READ_EXTERNAL_STORAGE` should be necessary.

vebde:
Hello,

My app clicks photo and sends it directly to the server without saving on device. 

How can I allow user to crop the image without using WRITE_EXTERNAL_STORAGE permission?

Please help.

osrl:
Even READ_EXTERNAL_STORAGE is not needed to read get image from gallery. Gallery app gives your app a URI base permission for only selected medias.

PR #985, opened 2019-04-16
Remove WRITE_EXTERNAL_STORAGE permission check for internal storage access

mkremer-jemmic:
This pull request removes the need of the _WRITE_EXTERNAL_STORAGE_ permission for most of this module's methods on Android. As most of the functions only work with the app's internal storage, they do not need any additional permission. This addresses the issue #905.

Overview of the changes:
- The _clean_ and _openPicker_ methods does no longer check for the _WRITE_EXTERNAL_STORAGE_ permission as it only clears a folder from the internal cache
- The _openCamera_ method does no longer require the _WRITE_EXTERNAL_STORAGE_ permission. If this permission is granted, photos and videos taken will be stored to the external storage as they have been until now. Otherwise, they will be temporarily stored in the app's internal cache storage (optional behavior).

### DIFF
--- a/android/src/main/java/com/reactnative/ivpusic/imagepicker/PickerModule.java
+++ b/android/src/main/java/com/reactnative/ivpusic/imagepicker/PickerModule.java
@@ -154,30 +154,22 @@ class PickerModule extends ReactContextBaseJavaModule implements ActivityEventLi
     public void clean(final Promise promise) {
 
         final Activity activity = getCurrentActivity();
-        final PickerModule module = this;
 
         if (activity == null) {
             promise.reject(E_ACTIVITY_DOES_NOT_EXIST, "Activity doesn't exist");
             return;
         }
 
-        permissionsCheck(activity, promise, Collections.singletonList(Manifest.permission.WRITE_EXTERNAL_STORAGE), new Callable<Void>() {
-            @Override
-            public Void call() {
-                try {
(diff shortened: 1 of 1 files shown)


danidfm:
Yes please, this is an unnecessary permissions that could affect acceptance from the users. This PR is mandatory.

ivpusic:
when running example application and trying to select image, I get permission deined. could you take a look into it?

mkremer-jemmic:
I noticed this too, although it worked at one point. I will have a look into this

mkremer-jemmic:
The error occurred because the Compression class tried to save (non-)resized images to an external path. I fixed this by checking whether or not the WRITE_EXTERNAL_STORAGE permission has been granted. It follows the same logic as in my first commit for this PR.
Also, I added the cache path to the file provider paths (provider_paths.xml) as taking pictures with the camera did otherwise not work for me in the example project.

gabgagnon:
@mkremer-jemmic So with your PR, we no longer upload the file on the External Storage, right?

osrl:
Hello, I think even READ_EXTERNAL_STORAGE is not needed to request media from gallery app. Media provider(gallery app) already gives URI based permission for only the requested medias. Please watch this talk from Ian Lake https://youtu.be/C28pvd2plBA?t=1472

PR #987, opened 2019-04-18
Added config instructions for android

muhammadsarim555:


ivpusic:
this should be automatically added by react-native link

Issue #994, opened 2019-04-21
Cancel at cropper cause error


fergusko:
### Version
Tell us which versions you are using: 

- react-native-image-crop-picker v0.23.1
- react-native v0.59

### Platform
- iOS

### Expected behaviour
openPicker with Cropper -> select image -> cancel at cropper -> select another image -> confirm at copper -> an image should be returned.

### Actual behaviour
openPicker with Cropper -> select image -> cancel at cropper -> select another image -> confirm at copper -> no image is returned.

Turn on debug mode show:
The callback openPicker() exists in module ImageCropPicker; but only one callback my be registered to a function in a native module.

### Steps to reproduce

1. openPicker with Cropper,  select image , cancel at cropper 

2. select another image

3. confirm at copper 

// stacktrace or any other useful debug info

Love react-native-image-crop-picker? Please consider supporting our collective:
👉  https://opencollective.com/react-native-image-crop-picker/donate

fergusko:
Possible Fix ?

ImageCropPicker.m

- (void)imageCropViewControllerDidCancelCrop:
(RSKImageCropViewController *)controller {
    [self dismissCropper:controller selectionDone:NO completion:[self waitAnimationEnd:^{
        if (**_self.currentSelectionMode != PICKER &&_** (self.currentSelectionMode == CROPPING  || [[self.options objectForKey:@"cropping"] boolValue])) {
            self.reject(ERROR_PICKER_CANCEL_KEY, ERROR_PICKER_CANCEL_MSG, nil);
        }
    }]];
}

ivpusic:
fixed in https://github.com/ivpusic/react-native-image-crop-picker/releases/tag/v0.24.1

PR #1000, opened 2019-05-02
iOS Bug Fix - Cancel at cropper cause error 

paul-hart:
https://github.com/ivpusic/react-native-image-crop-picker/issues/994

Fix suggested in the thread above.


mayconmesquita:
+1

ivpusic:
thanks

PR #1025, opened 2019-06-04
Fix bug where sometime Picture folder is not available on the device yet

Kishimotovn:


ivpusic:
merged here https://github.com/ivpusic/react-native-image-crop-picker/pull/1027

PR #869, opened 2018-11-03
Avoid conflicts with other SDKs that are using FileProvider

phips28:
I had a conflict with another SDK: https://github.com/kinecosystem/kin-ecosystem-android-sdk/issues/175

solution found here: 
https://github.com/stkent/bugshaker-android/issues/108#issuecomment-293291649

PR #1053, opened 2019-07-10
Fixing multiple file provider issue

jinxac:
I was using `react-native-image-crop-picker` in my project. In this project, the file provider is mentioned in the `AndroidManifest.xml` of the library itself.

In my project, I am using another library (`react-native-freshchat`) which also needs FileProvider.  Freshchat(or in future may be another library) requires the provider to be added in the application's `AndroidManifest.xml`. 

Because of multiple FileProviders, the build compilation was throwing an error and asking me to override the configuration in the application's `AndroidManifest.xml` using tools. I tried the same and compilation went fine but my `react-native-image-picker` openCamera stopped working.

So, this pull request takes care of such scenarios where we just extend the base FileProvider class. This is inspired by https://github.com/stkent/bugshaker-android/commit/5155a80f73406c4c3773a3e6873658b5c82a363b

### DIFF
--- a/android/src/main/AndroidManifest.xml
+++ b/android/src/main/AndroidManifest.xml
@@ -6,7 +6,7 @@
     <application>
 
         <provider
-            android:name="android.support.v4.content.FileProvider"
+            android:name=".FileProvider"
             android:authorities="${applicationId}.provider"
             android:exported="false"
             android:grantUriPermissions="true">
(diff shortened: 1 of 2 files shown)


jinxac:
Looks like PR is already created in https://github.com/ivpusic/react-native-image-crop-picker/pull/869/files. Closing this

PR #1054, opened 2019-07-10
Added tintColor property for the iOS options for openPicker function

nadav2051:
I've added this since I needed the possibility to change the font color of the text on the ImagePicker view.


nadav2051:
Any news on that PR?

Issue #1044, opened 2019-06-25
compressImageMaxWidth/compressImageMaxHeight not respected on iOS


howg0924:
### Version
Tell us which versions you are using: 

- react-native-image-crop-picker v0.24.1
- react-native v0.55.4

### Platform
Tell us to which platform this issue is related

iOS

### Expected behaviour

when setting both compressImageMaxWidth = compressImageMaxHeight = someValue;
the width and height of the result image should both smaller than someValue

### Actual behaviour

the width and height of the result image may be larger than someValue
the resize result may be wrong

### Steps to reproduce

1.set both compressImageMaxWidth and compressImageMaxHeight to 1000

2.open image picker

3.select an landscape image. For example, with width = 4000, height = 2000.

4.result image will have width = 2000, height = 1000

### Attachments

The bug is caused by the following code in Compress.m

```
if (maxWidth < maxHeight) {
    newWidth = maxWidth;
    newHeight = (oldHeight / oldWidth) * newWidth;
} else {
    newHeight = maxHeight;
    newWidth = (oldWidth / oldHeight) * newHeight;
}
```

if I change the code to:

```
CGFloat scaleFactor = ((maxWidth / oldWidth) < (maxHeight / oldHeight)) ? (maxWidth / oldWidth) : (maxHeight / oldHeight);

int newWidth = oldWidth * scaleFactor;
int newHeight = oldHeight * scaleFactor;
```

then the result is correct.

wootwoot1234:
@ivpusic I just submitted a simple bug fix pull request using @howg0924's code for this issue.  Here's the pull request: https://github.com/ivpusic/react-native-image-crop-picker/pull/1287.  I tested it and it's working.  Can you take a look and merge?

PR #1062, opened 2019-07-14
Add compressImageMaxPixels feature, unit tests

twelve17:
I wanted to be able to scale the image dimensions with a constraint in total pixels as opposed to a maximum width or height, as I was running into fairly wide images, which would end up a bit smaller than desired using the max height/width approach.

If a `compressImageMaxPixels` option is provided and is > 0, it takes precedence over `compressImageMaxWidth/compressImageMaxHeight`, and the latter two are ignored.

...

I had a hard time figuring out how to set things up for development of this library, so I took some liberty to guess at the opinion of how it should be set up.  Please let me know if I should change this.

Because I was adding complexity to the scaling portion of the compression code, I pulled out the scaling specific logic into its own file, and also added unit tests for both iOS and Android.  

In the process, I ran into this bug, so I fixed it too: https://github.com/ivpusic/react-native-image-crop-picker/issues/1044

Additional info:

- I could not figure out how to build the library in XCode.  There was no scheme for imageCropPicker.  I was afraid to break the compilation of the library from other projects, so I added the unit tests inside the example app, by adding a exampleTests target.

- I could not get Android to build using the version of gradle in the project, so I upgraded it to the latest, and configured `build.gradle`.  I also added `react-native` as a `devDependency` so I could refer to it from the `build.gradle` configuration.  Seeing that the XCode build looks for react-native in `../../react-native`, it is not consistent for the android equivalent to use a `devDependency`-based react-native from the `node_modules` installation, but I couldn't figure out how to build it otherwise.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,6 +38,7 @@ local.properties
 #
 node_modules/
 npm-debug.log
+package-lock.json
 
 # BUCK
 buck-out/
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ ImagePicker.openPicker({
 **Android: The prop 'cropping' has been known to cause videos not to be display in the gallery on Android. Please do not set cropping to true when selecting videos.**
 
 
-### Select from camera 
+### Select from camera
 
 #### Image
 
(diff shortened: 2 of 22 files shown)


dphaener:
@ivpusic What is the status of this? I'm using this forked version and am needing the RN 0.60 upgrades as well. Can we get this merged?

dphaener:
@ivpusic Ping on this. Any movement in getting this merged?

twelve17:
Closing in favor of re-based 1268

Issue #953, opened 2019-03-12
The order of the photos turned upside down in the photo list is returned


000xuandu:
### Version
Tell us which versions you are using: 

- react-native-image-crop-picker v0.21.3
- react-native v0.55.4

Looks like the order of the images in the library list is not correct with the order in which the photos are selected in the photo gallery.

Someone please give me a useful solution.

ivpusic:
order of photos is not guaranteed.

ssinss:
Check this. https://github.com/questbeat/QBImagePicker/issues/185#issuecomment-283101841

arpitjacob:
can we merge this fix that @ssinss has pointed out.

Issue #1087, opened 2019-07-31
Camera roll gallery opens to oldest images (iOS)


joshallison:
Hello,

when opening the camera roll gallery, the scroll position is automatically set to the top of the gallery (oldest images / videos).

Is there a way to open the camera roll gallery at the bottom (most recent) images like the iOS native photos app? 

Thank you so much for your help.

React native: 0.57.8
react-native-image-crop-picker: 0.24.1

<img width="464" alt="Screen Shot 2019-07-31 at 12 23 07 pm" src="https://user-images.githubusercontent.com/11746893/62178907-027d3a00-b38e-11e9-80f9-5d6b6f0630ef.png">

jhondavila:
I have the same problem brother, I have also published the same

https://github.com/ivpusic/react-native-image-crop-picker/issues/1089

migue1s:
I created a newly merged PR that allows you to modify the sort order, where you could sort the photos by creation date (newest first), would this help you @joshallison?

Note: this code is on master, it's not on any release yet

https://github.com/ivpusic/react-native-image-crop-picker/pull/1090

jhondavila:
> Creé un RP recién fusionado que te permite modificar el orden de clasificación, donde puedes ordenar las fotos por fecha de creación (la más nueva primero), ¿te ayudaría esto @joshallison ?
> 
> Nota: este código está en el maestro, aún no está en ninguna versión
> 
> [# 1090](https://github.com/ivpusic/react-native-image-crop-picker/pull/1090)

thank you,

jhondavila:
Could you show us an example of how to use it? we have tried using the new sortOrder configuration and we have not been successful with the order of the images


`ImagePicker.openPicker({
                compressImageQuality: 0.7,
                compressImageMaxWidth: 1200,
                mediaType: "photo",
                includeBase64: true,
                //smartAlbums: ["RecentlyAdded"]
                sortOrder : "asc"
            }).then(image => {
                this.props.onCaptureImage(
                    {
                        uri: image.path,
                        type: image.mime, // or photo.type
                        name: image.filename || `photo.jpg`
                    }
                );
                this.onCancel();
            }).catch((e) => {
                this.onCancel();
            });`

migue1s:
[...]

PR #1090, opened 2019-07-31
Adds a sortOrder flag to sort by creation date on iOS

migue1s:
Adds support for ordering the photos by creation date (ascending or descending) on iOS. By default, it'll keep the behavior that's currently on the library, which is that sort order is unspecified.

Related issues:
#953

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,6 +142,7 @@ ImagePicker.clean().then(() => {
 | mediaType                               |           string (default any)           | Accepted mediaType for image selection, can be one of: 'photo', 'video', or 'any' |
 | showsSelectedCount (ios only)           |           bool (default true)            | Whether to show the number of selected assets |
 | forceJpg (ios only)           |           bool (default false)            | Whether to convert photos to JPG. This will also convert any Live Photo into its JPG representation |
+| sortOrder (ios only)           |           string (default 'none', supported values: 'asc', 'desc', 'none')            | Applies a sort order on the creation date on how media is displayed within the albums/detail photo views when opening the image picker |
 | showCropGuidelines (android only)       |           bool (default true)            | Whether to show the 3x3 grid on top of the image during cropping |
 | showCropFrame (android only)       |           bool (default true)            | Whether to show crop frame during cropping |
 | hideBottomControls (android only)       |           bool (default false)           | Whether to display bottom controls       |
--- a/index.d.ts
+++ b/index.d.ts
@@ -28,6 +28,7 @@ declare module "react-native-image-crop-picker" {
         mediaType?: string;
         showsSelectedCount?: boolean;
         forceJpg?: boolean;
+        sortOrder?: 'none' | 'asc' | 'desc';
         showCropGuidelines?: boolean;
         hideBottomControls?: boolean;
         enableRotationGesture?: boolean;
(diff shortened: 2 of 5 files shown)


ivpusic:
thanks!

ivpusic:
@migue1s could you add support for this feature by using official latest version of QBImagePicker?

migue1s:
Got it, thanks for the heads up @ivpusic . I'll do it in a few days probably.

migue1s:
@ivpusic It seems like someone already did this on the official library on https://github.com/questbeat/QBImagePicker/pull/195 back in 2017. However, that lib hasn't been updated for 4 years and seems stale, does it still make sense to do the change on the original lib even when the maintainers seem to not be active?

danilobuerger:
@ivpusic QBImagePicker is not maintained anymore, I don't see how this can be added there.

migue1s:
Echoing what Danilo said, any way to go forward with this considering that the QBImagePicker isn't being maintained @ivpusic ?

ivpusic:
@migue1s could you reopen PR

migue1s:
It doesn't work as expected anymore, due to the dependency being source from cocoapods now instead of using the embedded one (from the RN 60 changes). I'll contact the folks over at QuestBeat and see if we can get a couple of PRs merged in there + submit a couple fixes afterwards

migue1s:
@ivpusic QuestBeat is unfortunately not responding :( I think the only paths forward are:

in the short term people can use https://github.com/ivpusic/react-native-image-crop-picker/issues/1087#issuecomment-527522579 to fix their environments

in the long term we should fork + republish the lib to make it accessible via cocoapods or completely deintegrate the lib from the cocoapod dependencies and rely on the one inside the repo only (I tried doing a PR with the latter solution but I got stuck with some linker issues).

ivpusic:
added in version 0.27.0 https://github.com/ivpusic/react-native-image-crop-picker/releases/tag/v0.27.0

cspicuzza:
Thanks for adding this feature.  Unfortunately it's not building with RN 0.6.5

```
The following build commands failed:
	CompileC {**projectpath**}/ios/build/Carbon/Build/Intermediates.noindex/Pods.build/Debug-iphonesimulator/RNImageCropPicker.build/Objects-normal/x86_64/ImageCropPicker.o {**projectpath**}/node_modules/react-native-image-crop-picker/ios/src/ImageCropPicker.m normal x86_64 objective-c com.apple.compilers.llvm.clang.1_0.compiler
(1 failure)
```

in attempts to troubleshoot I tried:
- removing/installing node_modules
- removing ios/build

ivpusic:
you have to add 

```pod 'QBImagePickerController', :path => '../node_modules/react-native-image-crop-picker/ios/QBImagePicker/QBImagePickerController.podspec'
```

to your Podfile and after that run `pod install`

Gavsum:
> you have to add
> 
> ```
> pod 'QBImagePickerController', :path => '../node_modules/react-native-image-crop-picker/ios/QBImagePicker/QBImagePickerController.podspec'
> ```
> 
> to your Podfile and after that run `pod install`

This is causing `No known class method for selector 'labelColor'` on build, any ideas ?

migue1s:
I have a feeling that your issue might be related to https://github.com/ivpusic/react-native-image-crop-picker/pull/1175 @Gavsum (haven't tried it yet though!)

Issue #957, opened 2019-03-14
openCropper, error [permission denied]


jaekwangLee:
### Version
Tell us which versions you are using: 

- react-native-image-crop-picker v0.23.0
- react-native v0.58.3

### Platform
Tell us to which platform this issue is related

- Android

### Expected behaviour

I hope to work like this...

1. User have to agree with WRITE_EXTERNAL_PERMISSION (<---in my code, not image-crop-picker)
2. User select any image in my flatlist
3. openCropper work
4. return, cropped image info

### Actual behaviour

1, 2, 3 is OK. very good.
But, If i selected any image and then press the ok button,
return error ...

Error: /storage/emulated/0/Pictures/new_filename.jpg  (Permission denied)

### Attachments
Possible Unhandled Promise Rejection (id: 0):
Error: /storage/emulated/0/Pictures/9fc670cf-738e-4c61-8a57-ebaf44a1d0af.jpg (Permission denied)
createErrorFromErrorData@http://localhost:8081/index.delta?platform=android&dev=true&minify=false:2012:26
http://localhost:8081/index.delta?platform=android&dev=true&minify=false:1964:51
__invokeCallback@http://localhost:8081/index.delta?platform=android&dev=true&minify=false:2531:23
http://localhost:8081/index.delta?platform=android&dev=true&minify=false:2262:34
 __guard@http://localhost:8081/index.delta?platform=android&dev=true&minify=false:2435:15
invokeCallbackAndReturnFlushedQueue@http://localhost:8081/index.delta?platform=android&dev=true&minify=false:2261:21
invokeCallbackAndReturnFlushedQueue@[native code]
ReactInstanceManager.detachViewFromInstance()

Love react-native-image-crop-picker? Please consider supporting our collective:
👉  https://opencollective.com/react-native-image-crop-picker/donate

Obi1Kennoby:
Any updates about it? I have the same issue with react-native v0.60.4 and react-native-image-crop-picker v0.25.0

matamicen:
First of all, thanks guys for this library :)

We have the same error in RN 0.60.4 with v0.25.0: 
**'/storage/emulated/0/Pictures/8ec90c24-e914-4b1e-b787-b27425e1fca1.jpg (Permission denied)'**

Here is the code, it exit thru the CATCH sentence:
```
  ImagePicker.openCropper({
                    path: data.uri,
                    width: 1000,
                    height: 1000
                  }).then(image => {
[...]

PR #1095, opened 2019-08-15
check permission before startCropping. Fix of #957

Obi1Kennoby:
The `openCropper` function does not check `WRITE_EXTERNAL_STORAGE` permission like `openPicker` or `openCamera` do, so in this PR I checking it (and ask for grant if needed) before startCropping method called. It's is a fix of [#957](https://github.com/ivpusic/react-native-image-crop-picker/issues/957) issue


ivpusic:
thanks!

Issue #1112, opened 2019-09-19
Could not determine artifacts for com.github.yalantis:ucrop:2.2.2-native


snamstorm:
### Version
Tell us which versions you are using: 

- react-native-image-crop-picker v0.25.2 (Also tested on v0.25.0)
- react-native v0.60.3

### Platform
Tell us to which platform this issue is related

- Android

### Steps to reproduce

Had no issue building a release version a week ago (no library versions in our app have updated since) - tried yesterday and started to get the below error `Could not determine artifacts for com.github.yalantis:ucrop:2.2.2-native`.

Saw there were some issues about a year back relating to `yalantis:ucrop:2.2.2-native` and `react-native-image-crop-picker`.

Seems that `ucrop` was migrated to AndroidX 3 days ago, not sure if that is causing this error with `react-native-image-crop-picker`
https://github.com/Yalantis/uCrop/releases

iOS build still works without any issue

### Attachments

Build Error:
```
FAILURE: Build failed with an exception.
* What went wrong:
Could not determine the dependencies of task ':app:compileDevReleaseJavaWithJavac'.
> Could not resolve all dependencies for configuration ':app:devReleaseRuntimeClasspath'.
   > Could not determine artifacts for com.github.yalantis:ucrop:2.2.2-native: Skipped due to earlier error
```

build.gradle
```
allprojects {
    repositories {
        // Add jitpack repository (added by react-native-spinkit)
        maven { url "https://jitpack.io" }
        mavenLocal()
        google()
        jcenter()
        maven {
            // All of React Native (JS, Obj-C sources, Android binaries) is installed from npm
            url "$rootDir/../node_modules/react-native/android"
        }
        maven {
            url "https://maven.google.com"
        }
        maven {
            // Android JSC is installed from npm
            url("$rootDir/../node_modules/jsc-android/dist")
        }
    }
}
```

```
buildscript {
    ext {
        buildToolsVersion = "28.0.3"
        minSdkVersion = 16
        compileSdkVersion = 28
        targetSdkVersion = 28
        supportLibVersion = "28.0.0"
	googlePlayServicesAuthVersion = "16.0.1" 
    }
    repositories {
        google()
        jcenter()
        maven { url 'https://maven.fabric.io/public' }
    }
    dependencies {
        classpath 'com.android.tools.build:gradle:3.4.1'
        classpath 'com.google.gms:google-services:4.2.0'
    }
}
```

harguus:
[...]

PR #1115, opened 2019-09-20
fixed `jitpack.io` url in readme.md

yurkagon:
related issue:
https://github.com/ivpusic/react-native-image-crop-picker/issues/1112


ivpusic:
hmm, based on an example which can be found on https://jitpack.io/ it should be as it is

ivpusic:
people are reporting that without `www` prefix does not work anymore for them, will merge PR. thanks @yurkagon

PR #1144, opened 2019-10-15
🤖 Add 0.25.0 android x 対応

nitaking:


ivpusic:
this should be included already now

PR #1146, opened 2019-10-16
Fix xCode warnings about main thread execution

Strate:
This kind of warnings also presents in output console if Main Thread Checker is active.

image


ivpusic:
thanks

PR #1174, opened 2019-11-16
CocoaPods - Properly Interpolate Tag Version

ptlu:
I faced an issue today where CocoaPods could not properly install this library because the tag version was incorrect.  This should fix that!

Why?
* Ruby uses `"` quotes when interpolation is required.  Using `'` will mean `v#{version}` 
  is used directly rather than providing the version number.

Changes:
1. User Ruby string interpolation when setting the tag version for the source.


ivpusic:
thanks

PR #1207, opened 2020-01-10
Update react-native cli dependency.

twelve17:
CocoaPods install in the example app does not work:

```
$ pod install

Command `config` unrecognized. Make sure that you have run `npm install` and that you are inside a react-native project.

[!] Invalid `Podfile` file: 785: unexpected token at ''.

 #  from /foo/react-native-image-crop-picker/example/ios/Podfile:37
 #  -------------------------------------------
 #  
 >    use_native_modules!
 #  end
 #  -------------------------------------------
```
This is because the example app's `package.json` points to react-native CLI v2, which has since been moved to `@react-native-community/cli` V3.

### DIFF
--- a/example/ios/Podfile.lock
+++ b/example/ios/Podfile.lock
@@ -223,7 +223,7 @@ PODS:
     - React-cxxreact (= 0.61.2)
     - React-jsi (= 0.61.2)
     - ReactCommon/jscallinvoker (= 0.61.2)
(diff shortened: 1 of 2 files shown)


ivpusic:
thanks! these changes are pushed as part of recent commits, so will close this one. tnx for pointing out on this issue!

PR #1216, opened 2020-01-20
Add imageOutput in Android to fix png layer

KarlosQ:
When you try to crop a png, you lose the transparency layer. This PR fix that problem in Android.

Cropping a png in master:
bad

Cropping a png with this PR:
good


KarlosQ:
I will create other PR with iOS too

PR #1221, opened 2020-01-30
Image output format for keep transparent layer in png

KarlosQ:
When you try to crop a png, you lose the transparency layer. This PR fix that problem.

Cropping a png in master:
bad

Cropping a png with this PR:
good


KarlosQ:
I will create other PR for a better solution

PR #1254, opened 2020-03-25
Add support for newer versions of Android with image picking

rogerkerse:
When selecting an image, the file path came like "file://somepath" but newer versions of android require "content://somepath".

If that is not the case, this image cannot be used to share it outside of the app.
This pull request prevents "Exposed beyond app" error.


muganwas:
Even with this change I somehow can't get the image uploaded to an external server.

Issue #1263, opened 2020-04-07
Folder to save image taking from camera


quyenlt1812:
### Version
- react-native-image-crop-picker v0.26.1
- react-native v0.61.4

### Platform
- Android

I really want to have my own folder (custom folder instead save to Pictures, maybe sub folder in Pictures) to save image has been taking from the camera so I can easily find it in the gallery.

PR #1264, opened 2020-04-07
Android pictures sub-folder

quyenlt1812:
Just create a custom folder in Pictures so the images is easier to manage

Related issues:
https://github.com/ivpusic/react-native-image-crop-picker/issues/1263


VincentCATILLON:
@ivpusic what about this PR ? :)

PR #1265, opened 2020-04-08
Update README.md (Clarify avoidEmptySpaceAroundImage option)

MujtabaFR:
Clarify that **avoidEmptySpaceAroundImage** option is (**ios only**)


ivpusic:
thanks

Issue #1113, opened 2019-09-19
Gif files are returned as jpeg


fega:
### Version
Tell us which versions you are using: 

- react-native-image-crop-picker v0.25.2
- react-native v0.59.8

### Platform
IOS
### Expected behaviour
I'd expect as mentioned in previous comments that if I set compressImageQuality as undefined it will return original image but it doesn't work (I tried also with values like 1, 100 and null, but i'm getting the same)

### Actual behaviour

image is a JPEG

### Steps to reproduce
```
const options= {}
ImagePicker.openPicker(options).then(response => {
        console.log('MIME TYPE= ', response.mime)
});
```
### Attachments

// stacktrace or any other useful debug info

Love react-native-image-crop-picker? Please consider supporting our collective:
👉  https://opencollective.com/react-native-image-crop-picker/donate

brandonpearcy:
I can also confirm this is happening. If a user selects an animated GIF, the result is a static JPEG instead.

Adarwall:
Did you find a solution? I'm facing the same issue...

brunowbbs:
Did you find a solution?

STIENLILLY:
Same issue on RN 0.62.2

STIENLILLY:
https://github.com/ivpusic/react-native-image-crop-picker/blob/da9af08384c35dcd3be154cdb6d5c579b48b338a/ios/src/ImageCropPicker.m#L828-L841

'filePath' is forced to jpg extension that is created by the 'writeTempFile' option.

STIENLILLY:
https://github.com/ivpusic/react-native-image-crop-picker/pull/1301 Check for my PR.

omarinho:
Apparently the same is happening with .BPM extension,

STIENLILLY:
@omarinho .bpm is not support to extension but you can customize persistFile function to support bpm

PointCodeZero:
I saw that a fix was merged here [#1301 ](https://github.com/ivpusic/react-native-image-crop-picker/pull/1301). Any plans on publishing this fix on v.0.25.3 for RN 0.60? It would be a saviour for us!

LeonelRiera:
Any news with this fix? It seems correct but why has it not been merged after years? .gif is still not working and it is still returning a .jpg

PR #792, opened 2018-08-14
Add heic format for determining MIME type

s-lee:
HEIC format (introduced in iOS 11) needs to be handled for determining MIME type

PR #1301, opened 2020-06-02
(ios) Fix to writeTempFile .gif

STIENLILLY:
https://github.com/ivpusic/react-native-image-crop-picker/issues/1113 Gif files are returned as jpeg

### DIFF
--- a/ios/src/ImageCropPicker.m
+++ b/ios/src/ImageCropPicker.m
@@ -829,8 +829,16 @@ RCT_EXPORT_METHOD(openCropper:(NSDictionary *)options
     // create temp file
     NSString *tmpDirFullPath = [self getTmpDirectory];
     NSString *filePath = [tmpDirFullPath stringByAppendingString:[[NSUUID UUID] UUIDString]];
-    filePath = [filePath stringByAppendingString:@".jpg"];
-
+    NSString *mimeType = [self determineMimeTypeFromImageData:data];
+    if([mimeType isEqualToString:@"image/jpeg"]){
+     filePath = [filePath stringByAppendingString:@".jpg"];
+    }
+    else if([mimeType isEqualToString:@"image/gif"]){
+     filePath = [filePath stringByAppendingString:@".gif"];
+    }
+    else {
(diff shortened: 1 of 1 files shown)


abdelmagied94:
I think we can make a function for getting image extension from the data (same as the mime one)
```swift
- (NSString *)determineExtensionFromImageData:(NSData *)data {
    uint8_t c;
    [data getBytes:&c length:1];

    switch (c) {
        case 0xFF:
            return @".jpg";
        case 0x89:
            return @".png";
        case 0x47:
            return @".gif";
        case 0x49:
        case 0x4D:
            return @".tiff";
    }
    return @".jpg";
}
```

STIENLILLY:
@oferRounds @abdelmagied94 Thank you for reviews of code. I added a code for checking file extensions determined by data.

bitcrumb:
This is good fix. No idea why the original implemenation resorts to just adding ".jpg"?
Anyway, maybe also a good idea to add mime for heic to the list, see: https://github.com/ivpusic/react-native-image-crop-picker/pull/792/files

STIENLILLY:
@bitcrumb Thank you for advising me. I couldn't find the reason why the original implementation is forced to jpg extensions although the default value of the `forceJpg` property is `false`.
And, I revamped the logic detecting the HEIF file format because the null value(0x00), the first of the HEIF magic number, is improper to detect the HEIF file format.

bitcrumb:
@STIENLILLY Thanks for your feedback! I was not aware (I am not deeply invested into this matter). I will apply the same fix to the patches in our codebase.

FYI: we are using https://github.com/ds300/patch-package to apply patches to react native libraries, especially for repositories like this where there is little to no feedback on useful PRs like yours.

Thanks for your efforts!

samerdernaika:
its not working

junqdu:
its not working

belal-bh:
I was trying to crop `png` file and getting output as `jpg`. Is there any way to get the same type of image as input image?

changwoolab:
Why isn't this merged yet?

Issue #181, opened 2017-01-02
Is there any possible to return duration of the picked video?


jaggerwang:
I want to limit the duration of video picked by our app's user to short than 30s. At now, the props of object returned by picker including width, height, mime and size, but no duration. And I don't want to do this check on the server side.

ivpusic:
currently this is not possible, but this would be definitely very useful. If you have some time I would accept PR with this feature. My proposal would be to include new property called `metadata`. Here we would put all extra data we know about asset. Asset can be image or video. For images later we can include some image specific information, and for videos as well. Let me know what do you think.

jaggerwang:
I'd like to help, but I have no iOS or Android developing experience, my most programming experience is on server side. I'm sorry for this. Hoping some nice guy will add this feature later.

ivpusic:
No problem. Will close this issue for now, but will keep it in mind.

jpmazza:
+1

BaderSerhan:
did we find a solution for this? i want to implement a feature for uploading a video from library and trimming it, but i am not finding a good solution so far...

astraeapark:
+1

Lunziyuan:
+1

Coder108:
When will this feature appear?

kostiantyn-solianyk:
@ivpusic @astraeapark @Lunziyuan @Coder108 
Maybe it's too late. But I also need this functionality.
So, I've made this issue working on both **ANDROID and IOS**. I hope @ivpusic will look into PR that I sent. https://github.com/ivpusic/react-native-image-crop-picker/compare/master...kostiantyn-solianyk:master

ivpusic:
available in https://github.com/ivpusic/react-native-image-crop-picker/releases/tag/v0.32.2

PR #1325, opened 2020-06-26
Duration of the picked video (in milliseconds)

kostiantyn-solianyk:
PR for possibility to get duration of the picked video.
related issue - https://github.com/ivpusic/react-native-image-crop-picker/issues/181

This PR will be helpful for video validation, especially when you need to set a limit for uploaded video.

**Implemented on both (Android | IOS)**
In **"Response Object"** you will get duration property in milliseconds.

Tested on Android and IOS.

Duration will be available from local file(path) or cloud(any link as well)

Updated READMY file => **Response Object section**

The PR doesn't have additional dependencies or global changes.

### DIFF
--- a/ios/src/ImageCropPicker.m
+++ b/ios/src/ImageCropPicker.m
@@ -434,6 +434,10 @@ RCT_EXPORT_METHOD(openCropper:(NSDictionary *)options
             [outputURL getResourceValue:&fileSizeValue
                                  forKey:NSURLFileSizeKey
                                   error:nil];
+            
+            AVURLAsset *durationFromUrl = [AVURLAsset assetWithURL:outputURL];
+            CMTime time = [durationFromUrl duration];
+            int seconds = ceil(time.value/time.timescale) * 1000;
 
             completion([self createAttachmentResponse:[outputURL absoluteString]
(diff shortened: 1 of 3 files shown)


ios/src/ImageCropPicker.m
```diff
@@ -434,6 +434,10 @@ - (void) handleVideo:(AVAsset*)asset withFileName:(NSString*)fileName withLocalI
             [outputURL getResourceValue:&fileSizeValue
                                  forKey:NSURLFileSizeKey
                                   error:nil];
+            
+            AVURLAsset *durationFromUrl = [AVURLAsset assetWithURL:outputURL];
+            CMTime time = [durationFromUrl duration];
+            int seconds = ceil(time.value/time.timescale) * 1000;
```
ivpusic, line 440: I guess name of variable here should be milliseconds?

kostiantyn-solianyk: oh, yes, you are right

kostiantyn-solianyk: Changed

ivpusic:
thanks a lot!

PR #1332, opened 2020-07-04
Add types for smart albums, compressed video presets, and picker errors

dburdan:
Several simple type improvements.


ivpusic:
thanks a lot!

PR #1333, opened 2020-07-04
Conditional return types, and documentation for every property

dburdan:
This PR covers a few type-related improvements.

- Separate `Image` and `Video` types.
- Make return types for `openPicker` and `openCamera` conditional. Now the return types change based on the input props. For example, the return type will only be an array if the `multiple` prop is present.
- Add documentation to every property for improved IntelliSense.

I've used this in production for several weeks but another set of eyes would be awesome.


ivpusic:
thanks!

PR #1354, opened 2020-07-24
Fixes possibly nil window lookup

ashfurrow:
Good morning! This is a really cool package, but I faced a slight setback when trying to integrate it into [our application](https://github.com/artsy/eigen). The problem lies with some swizzling we do with our application delegate (using [JSDecoupledAppDelegate](https://github.com/JaviSoto/JSDecoupledAppDelegate)). It's not a common practice, especially among React Native applications, but the result is that `[[[UIApplication sharedApplication] delegate] window]` returns `nil`. This causes the call to [`presentViewController:animated:completion:`](https://github.com/ivpusic/react-native-image-crop-picker/blob/a0c7a8f4b591c864c8025d0eb2360b8270311301/ios/src/ImageCropPicker.m#L360) to silently fail. In our app, we typically use `[[UIApplication sharedApplication] keyWindow]` instead, although this was [deprecated in iOS 13](https://developer.apple.com/documentation/uikit/uiapplication/1622924-keywindow). 

This PR changes the library to access the [application's `windows` property](https://developer.apple.com/documentation/uikit/uiapplication/1623104-windows). This is probably a bit more accurate to what you want, since it handles multi-window applications on iPad, and the ordering of the array matches the visual ordering for the user. This means that the last window in the array is always the top-most window that the user most-recently interacted with, which is probably where the modal image picker should be presented.

Happy to clarify anything, and thanks again!

### DIFF
--- a/ios/src/ImageCropPicker.m
+++ b/ios/src/ImageCropPicker.m
@@ -124,7 +124,7 @@ RCT_EXPORT_MODULE();
 }
 
 - (UIViewController*) getRootVC {
-    UIViewController *root = [[[[UIApplication sharedApplication] delegate] window] rootViewController];
+    UIViewController *root = [[[[UIApplication sharedApplication] windows] lastObject] rootViewController];
     while (root.presentedViewController != nil) {
         root = root.presentedViewController;
     }


ivpusic:
thanks!

mrousavy:
@ivpusic may I ask why you reverted this?

ivpusic:
because of this issue https://github.com/ivpusic/react-native-image-crop-picker/issues/1383. if you have some time to check how we could resolve it, would be cool

ashfurrow:
Hmmm, sorry to have caused issues for the project! Seems like our app setup is even weirder than I thought – though we are trying to move it closer to RN standards. Would you be open to performing both lookups and using whichever one returns an instance? I’m also happy to continue using patch-package to make this work for our app.

ivpusic:
would say for now let’s keep it as it is. if more people are having similar use case, than we could apply this again

PR #1357, opened 2020-07-28
fix: fix error for unknown class method for selector 'LabelColor' in …

Preeternal:
Fix for the [issue](https://github.com/ivpusic/react-native-image-crop-picker/issues/1275)


Preeternal:
closed as inactive (probably we don't need XCode 10 any more)

Issue #1293, opened 2020-05-27
sourceURL is always null in ios


Rishabh-spec:
### Version
Tell us which versions you are using: 

- react-native-image-crop-picker v0.31.1
- react-native v0.61.3

### Platform
Tell us to which platform this issue is related

- iOS

### Expected behaviour
sourceURL can not be null

### Actual behaviour
sourceURL getting null in ios

### Steps to reproduce

1. Open Image Picker

2. Select Multiple image

3. Click Done 

### Attachments

// stacktrace or any other useful debug info

Love react-native-image-crop-picker? Please consider supporting our collective:
👉  https://opencollective.com/react-native-image-crop-picker/donate

Jackychen8:
Same thing is happening for me now. It had worked as recently as June 14th, 2020 though. I had only started working with this library a few days before that.

hengkx:
how to resolve?

hengkx:
https://stackoverflow.com/questions/57202965/phimageresultisdegradedkey-phimagefileurlkey-is-not-found

hengkx:
#1130

ivpusic:
fixed in https://github.com/ivpusic/react-native-image-crop-picker/releases/tag/v0.33.3

Issue #1130, opened 2019-09-26
sourceURL always null on iOS 13 devices and simulators


mzdon:
### Version
Tell us which versions you are using: 

- react-native-image-crop-picker v0.25.2
- react-native v0.60.5

### Platform
Tell us to which platform this issue is related

iOS 13

### Expected behaviour

The response object in JS returns a sourceURL on iOS when images are selected.

### Actual behaviour

The response object sourceURL property is always null for selected images.

### Steps to reproduce

1. Open the image picker on any iOS 13 device or simulator.

2. Select one or more images (videos are not affected) and press Done.

3. Inspect the response objects in JS and observe that the sourceURL is always null.

### Attachments

It looks like this issue is related to the use of PHImageManager:requestImageDataForAsset which is listed as deprecated on the Apple Developer docs.  The NSDictionary info provided to the resultHandler callback used to contain a PHImageFileURLKey property but this property is no longer provided in iOS 13.  This property was set as the sourceURL.

It looks like there may be other ways to retrieve this info on iOS 13.  https://stackoverflow.com/a/57307250

I'm looking into this as well, but I have limited experience with Objective-C.

**EDIT:**
It seems this is only the case when the writeTempFile flag is false, which is the case for us.  Switching it back to true (the default) resolves that issue.  Maybe this is just a matter of updating documentation if this is an expected behavior?

maron-codersinteractive:
I have the same issue with `ImagePicker.openCamera()` but not with `ImagePicker.openPicker()`. I tested in version < iOS 13

boaz-lachman:
+1 have issue with openpicker

ArronJLinton:
For choosing from gallery and taking a picture, I was able to get through with a successful photo upload by using the path key in the response object.

kamilzamerski:
The same problem... SourceURL is null. Unable to get photo path

IvanCaceres:
using the path key `response.path` works on ios 13 for upload purposes

webbhead777:
You can use `path` to get the tmp file path but if you need to get the file path to the persisted photo (in the cameraroll) `sourceURL` is always `null` so it's not gonna work 😕

ivpusic:
fixed in https://github.com/ivpusic/react-native-image-crop-picker/releases/tag/v0.33.3

PR #1368, opened 2020-08-08
fix ios13 sourceUrl

hengkx:
ref #1130 #1293


ivpusic:
currently getting sourceURL only when selecting multiple files. could you add fix also for single selection mode?

hengkx:
done

omid-ebrahimi:
@ivpusic Please review this merge request. We strongly need it.

Issue #1290, opened 2020-05-13
Circle crop, causes image to show circle only. Cropped part is white.


bvelasquez:
### Version
Tell us which versions you are using: 

- react-native-image-crop-picker v0.31.1
- react-native v0.62.2

### Platform
Tell us to which platform this issue is related

- iOS
- Android

### Expected behaviour

Create a cropped square image given the width and height.

### Actual behaviour

![b4c7df49-c41d-11e3-a732-00142273155d](https://user-images.githubusercontent.com/2302210/81758732-85800480-9477-11ea-8bb2-4cb619461dc0.png)

Image is cropped as a square image, but with only a circle with the original image, rest is white background. It used to produce a square image without the white mask in the background.

### Steps to reproduce

1. use openPicker to crop an image with cropping: true, and cropperCircleOverlay: true.

2. 

3.

### Attachments

// stacktrace or any other useful debug info

Love react-native-image-crop-picker? Please consider supporting our collective:
👉  https://opencollective.com/react-native-image-crop-picker/donate

tastafur:
In my version in Android circle not crop

joshjens:
same issue here, expect a circular crop mask to still produce a square image. Adding the white frame matte can result in weirdness when using a borderRadius on a view to crop on display

skizzo:
+1, I'd also expect that `cropperCircleOverlay={true}` doesn't actually remove the content on the edges. Is there a way to achieve that? Thanks!

jcharbo:
Simply remove this method from imagecroppicker.m and you will get back a square.
`- (void)cropViewController:(TOCropViewController *)cropViewController didCropToCircularImage:(UIImage *)image withRect:(CGRect)cropRect angle:(NSInteger)angle {
    [self imageCropViewController:cropViewController didCropImage:image usingCropRect:cropRect];
}`

ivpusic:
fixed in https://github.com/ivpusic/react-native-image-crop-picker/releases/tag/v0.33.0

skizzo:
Amazing, thank you!

PR #1375, opened 2020-08-20
[iOS] Fixed to be cropped the same as Android

ifsnow:
Related to : #1290 

If we use `cropperCircleOverlay` in iOS, it's different from Android, it creates a circular cropped image on a white background. 

* **iOS**
wrong_image

* **Android**
correct_image

I think it's good to have the same result regardless of platforms. An image with a white background makes it difficult to use in a service by using `borderRadius`.

Please review positively.


ivpusic (approved):

ivpusic:
thanks!

PR #1384, opened 2020-08-26
Update ImageCropPicker.m

khurram-wasim:
This param can be used as

ImagePicker.openCamera({
  mediaType: 'video',
  durationLimit: LIMIT_IN_SECONDS
})


hannojg:
Great that you started working on it! 
This seems to be just the iOS implementation. How about the Android implementation? 

Also the typescript Definition would need to be updated as well as the documentation I guess 🤔

2xSamurai:
Hi. 
Is there a chance for getting this merged anytime soon?
Also will android also get a feature like this?
Thank you for the time and effort.

mrousavy:
I've started working on supporting `maximumVideoDuration` for iOS and Android, using `openCamera` as well as `openPicker` at #1463

PR #1387, opened 2020-08-27
Using both compressImageMaxWidth and compressImageMaxWidth fixed.

sungsong88:
When both compressImageMaxWidth and compressImageMaxHeight are set, only the compressImageMaxHeight was used.
Now it applies accordingly depends on images width is greater than the height or not.


sungsong88:
This is duplicate of https://github.com/ivpusic/react-native-image-crop-picker/pull/1391. Closing

PR #1391, opened 2020-08-30
Using both compressImageMaxWidth and compressImageMaxWidth fixed.

sungsong88:
When both compressImageMaxWidth and compressImageMaxHeight are set, it applies accordingly based on images width and height. Previously when both are set, only the compressImageMaxHeight was used.


orzhtml:
When can we merge?

anhnch:
Please merge!

PR #1394, opened 2020-09-04
[feature]: Add Cancel API

retyui:
Now users can use and know about cancel behavior:

```javascript
import ImagePicker, { isCancel } from "react-native-image-crop-picker";
ImagePicker.openPicker({
  mediaType: "photo"
}).then(() => {
  // ...
}).catch(error => {
  if (isCancel(error)) {
    // User cancelled the picker, exit any dialogs or menus and move on
  } else {
    throw error;
  }
});
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,26 @@ ImagePicker.clean().then(() => {
 });
 ```
 
+### Cancel piker or close camera
+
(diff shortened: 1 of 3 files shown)


ivpusic:
thanks for PR. let's keep utility functions out of this package for now

retyui:
@ivpusic I think we should add a note about an E_PICKER_CANCELLED error code in the README file

PR #1395, opened 2020-09-04
Add path field to ImageOptions on typescript

bonomoon:
When I used openCropper method, TypeScript says 'path' is not available field for parameter object type.
The parameter object type of openCropper method is ImageOptions. There is not path field on ImageOptions Type.

image


ivpusic (approved):

ivpusic:
thanks

PR #1421, opened 2020-10-01
[android] filter possible image types for cropping

oxyii:
Now `react-native-image-crop-picker` allow to choose any `image/*` file for cropping. But related to https://github.com/Yalantis/uCrop/issues/166#issuecomment-245606245 only `libjpeg` and `libpng` under the hood libUCrop. So I think will be useful to filter image files only ready for cropping

ref https://github.com/ivpusic/react-native-image-crop-picker/issues/1367#issuecomment-686576364

PR #1422, opened 2020-10-01
[android] filter image types if cropping only

oxyii:
@ivpusic Sorry, my bad. In PR #1421 I didn't think about those who don't need cropping, who need the picker only. Now images filtered if cropping only


ivpusic (approved):

ivpusic:
thanks!

PR #1439, opened 2020-10-17
Make list items accessible

v-miibr:
We noticed that one of the views for selecting images was not accessible and would like to make it accessible. This was based on user feedback.


shiweiwei114:
This is great! Currently the user is not able to select any photos in accessibility mode (VoiceOver), we need this! :)

FrozenIce0617:
You saved my day, @v-miibr 
It would be great if we can merge this change asap.

marcshilling:
You should also add:
`cell.accessibilityTraits = UIAccessibilityTraitImage | UIAccessibilityTraitButton;`

This will make it clear that it's an image that can be selected, and iOS will even sometimes be able to detect the contents of the image and read a description...super cool.

I'm using this patch and it's been working great. Should definitely be merged in with my suggested change!

pnikhilp:
Any update on when will be this change available in the package?

jan-happy:
This would be super helpful for a lot of people, please consider merging this!

Nanjundan (approved):

Nanjundan:
Hello @ivpusic Can you consider merging this PR with AccessibilityTraits?

PR #1287, opened 2020-05-06
Update Compression.m

wootwoot1234:
Fixes bug where compressImageMaxWidth and compressImageMaxHeight are not respected.

Credit to https://github.com/ivpusic/react-native-image-crop-picker/issues/1044

PR #995, opened 2019-04-26
Fix/android image resize

bonesyblue:
## Motivation
Whilst using this library, I discovered a bug with the image compression module on Android, which for my use case is resulting in correctly resized/compressed landscape images but incorrectly resizes/compresses portrait images.

As part of this PR, I have also updated the iOS compression module code to be consistent with my updates to the Android compression module. I have tested my changes on an Android and iOS device.

This PR provides a potential solution for the following issues:
https://github.com/ivpusic/react-native-image-crop-picker/issues/532
https://github.com/ivpusic/react-native-image-crop-picker/issues/908

## Detailed example:
Taking the current code in the compression module and using example values, the bug for my use case can be better understood:

Landscape image
```java
// maxWidth: 2048 (compressImageMaxWidth in the request options)
// maxHeight: 2048 (compressImageMaxHeight in the request options)
// Example landscape image with dimensions (w x h): 4096 x 2048

 File resize(String originalImagePath, int maxWidth, int maxHeight, int quality) throws IOException {
        Bitmap original = BitmapFactory.decodeFile(originalImagePath);

        int width = original.getWidth(); // 4096
        int height = original.getHeight(); // 2048

        float ratioBitmap = (float) width / (float) height; // evaluates to 2
        float ratioMax = (float) maxWidth / (float) maxHeight; // evaluates to 1!

        int finalWidth = maxWidth; // 2048
        int finalHeight = maxHeight; // 2048

        // IMHO I think we should be referencing the original image aspect ratio and not the aspect
        // ratio specified as part of the compression options. The compression options are only specifying 
        // the bounds within which we should be resizing but we should be concerned with respecting the 
        // original image aspect ratio here. In addition, the maxHeight and maxWidth parameters do not 
        // always respect the original image aspect ratio. Thus arbitrary values can be assigned and, as far 
        // as I understand, this would result in stretching the original image.
        if (ratioMax > 1) { // evaluates to false
            finalWidth = (int) ((float) maxHeight * ratioBitmap);
        } else {
            finalHeight = (int) ((float) maxWidth / ratioBitmap); // evaluates to 1024
        }

        // Creates a new scaled bitmap with dimensions: 2048 x 1024 (w x h) is ok ✅
        Bitmap resized = Bitmap.createScaledBitmap(original, finalWidth, finalHeight, true);
        ...
 ```

Portrait image
```java
// maxWidth: 2048 (compressImageMaxWidth in the request options)
// maxHeight: 2048 (compressImageMaxHeight in the request options)
[...]

PR #1457, opened 2020-11-08
Update resize function to respect both maxWidth and maxHeight on android and ios

OliverEvans96:
This is a problem that several others have pointed out and submitted PRs for, but I think my solution is slightly more elegant than any other I've seen ;) so I'll add it into the mix. Ultimately, though, it of course doesn't matter which one is accepted as long as the issue is resolved.

If `compressImageMaxWidth` and `compressImageMaxHeight` are both specified, they should both be respected. Currently, only the smaller one is respected. This is only a resonable approach for square images. In general, to satisfy both constraints, one can:
- Determine the scaling factors necessary to satisfy each constraint individually
- Apply the resizing that produces the smaller final image.

For example:
- An image is originally `2000 x 3000`
- I specify `maxWidth=500` and `maxHeight=600`
- The width constraint requires a scaling factor of `500 / 2000 = 0.25`
- The height constraint requires a scaling factor of `600 / 3000 = 0.20`
- Apply the smaller scaling factor of `0.20`, which represents the stronger constraint
- The final image is `400 x 600`. Both constraints are satisfied.

Related (there may be others):
#995 #1147 #1287 #1391 #1440 

Thanks!
Oliver

### DIFF
--- a/android/src/main/java/com/reactnative/ivpusic/imagepicker/Compression.java
+++ b/android/src/main/java/com/reactnative/ivpusic/imagepicker/Compression.java
@@ -41,17 +41,13 @@ class Compression {
         int rotationAngleInDegrees = getRotationInDegreesForOrientationTag(originalOrientation);
         rotationMatrix.postRotate(rotationAngleInDegrees);
 
-        float ratioBitmap = (float) width / (float) height;
-        float ratioMax = (float) maxWidth / (float) maxHeight;
+        float widthRatio = (float) maxWidth / (float) width;
+        float heightRatio = (float) maxHeight / (float) height;
+        // Resize the image by whichever ratio will make the image smaller, satisfying both constraints.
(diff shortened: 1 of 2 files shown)


Fischaela:
Would be great if this would get merged :)

OliverEvans96:
In the meantime, you might just have to [`patch-package`](https://www.npmjs.com/package/patch-package).

BaderSerhan:
Any updates on this?

anhnch:
> In the meantime, you might just have to [`patch-package`](https://www.npmjs.com/package/patch-package).

Thank you for the fix and patch-package hint! It's been 2 years, and the issue still remains :(

LCseniordev:
Hi, thanks for this great package!
Still can be reproduced in v0.41.1. Any updates?

PR #1500, opened 2020-12-26
Added useCroppedDimensions to iOS to keep the cropped rect size

AidenMontgomery:
We needed to be able to receive the largest possible image, within the bounds of the existing `compressImageMaxWidth` and `compressImageMaxHeight` settings. We found that if we did not set the `height` and `width` it was defaulting to 200, or we had to specify a larger `height` and `width`.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -316,6 +316,14 @@ declare module "react-native-image-crop-picker" {
          * @default Android: 1, iOS: 0.8
          */
         compressImageQuality?: number;
+
+        /**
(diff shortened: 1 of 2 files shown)


Maker-Mark:
Can this be reviewed by a maintainer? Known issue for 3 years 😢 
https://github.com/ivpusic/react-native-image-crop-picker/issues/2001

Ali-Kasasbeh (approved):

jasonroelofs:
Adding a +1. Five years later this patch still needs to be applied.

PR #1506, opened 2021-01-07
fix: Make sure to return precise + platform-consistent permission error code

### LINKED ISSUE #770: Inconsistent error codes
I've got some questions about the error codes that are used because I've just had to fix a "bug" in our app that was due to a different error code for missing permissions.

```javascript
ImagePicker.openCamera({...})
.then(...)
.catch(error => {
  if (error === "E_PERMISSION_MISSING") {
    // Do stuff...
  }
});


ImagePicker.openPicker({...})
.then(...)
.catch(error => {
  if (error === "E_PERMISSION_MISSING") {
    // Do stuff...
  }
});
```

Here's the methods I originally had and as you can see, I assumed that `E_PERMISSION_MISSING` was the code to use to check that the user hadn't previously denied a permission. It turns out that the correct code to use for `openCamera()` is `E_PICKER_NO_CAMERA_PERMISSION`.

Are the error codes documented anywhere? I checked the README and searched issues and couldn't find that information out without having to debug it by logging the error codes. That's easy enough to do but it seems like the kind of thing that should be documented somewhere.

Apologies if it is documented, I just haven't been able to find it yet...

AmauryLiet:
Hi!

First of all, thanks for the lib and your work on it 💪 

This PR brings consistency in permission-related error codes between iOS & android 
=> breaking change

Fixes #770 

PS: tested on iOS + android (except ios camera flow, due to simulator limitation)


AmauryLiet:
hey @ivpusic did you have the chance to get a look on the PR?

ivpusic:
thanks!

ivpusic:
available in 0.36.0

Issue #1415, opened 2020-09-26
Failed on select Video - IOS 14


RZulfikri:
### Version
Tell us which versions you are using: 

- react-native-image-crop-picker v0.35.0
- react-native v0.63.2

### Platform
Tell us to which platform this issue is related

- iOS 14

### Expected behaviour
Return video

### Actual behaviour
Return error

### Steps to reproduce

1. Open picker

2. Select video

3. Return error

```
code:E_CANNOT_PROCESS_VIDEO
message:Cannot process video data
domain:RCTErrorDomain
userInfo:null
```

### Attachments
When I'm trying to debug in xcode, exportSession always return status = 4 (AVAssetExportSessionStatusFailed), and the error look like this.

Video Export Failed: Error Domain=AVFoundationErrorDomain Code=-11800 "The operation could not be completed" UserInfo={NSUnderlyingError=0x2826f6730 {Error Domain=NSOSStatusErrorDomain Code=-16979 "(null)"}, NSLocalizedFailureReason=An unknown error occurred (-16979), NSURL=file:///var/mobile/Media/PhotoData/Metadata/DCIM/102APPLE/IMG_2272.medium.MP4, NSLocalizedDescription=The operation could not be completed}

[Update]
I'm trying to select few files:
1. file:///var/mobile/Media/PhotoData/Metadata/DCIM/102APPLE/IMG_2272.medium.MP4 (ERROR) // from whatsapp
2. file:///var/mobile/Media/DCIM/100APPLE/IMG_0454.MOV (CORRECT) // save from instagram
3. file:///var/mobile/Media/DCIM/102APPLE/IMG_2429.MOV (CORRECT) //  save from camera
4. file:///var/mobile/Media/PhotoData/Metadata/DCIM/102APPLE/IMG_2075.medium.MP4 (ERROR) // from whatsapp

Video with .MOV extension return success, but video with .MP4 return error.

Any idea why this issue exist and how to solve??

Love react-native-image-crop-picker? Please consider supporting our collective:
👉  https://opencollective.com/react-native-image-crop-picker/donate

vrgimael:
Same here, iOS 14 is causing a lot of trouble

chr-sk:
Also experiencing the same issue with iOS 14 and iOS 14.2 beta.

TONYYYYYYYYYYY:
It's related to downloading images from iCloud. 

[...]

PR #1515, opened 2021-01-20
Fix video not loading from iCloud

luco:
Fix the video not loading/downloading from iCloud as suggested by @anhkieet in #1415.


ivpusic:
thanks!

Jojr:
Thanks! NPM still with v0.35.2.

PR #1521, opened 2021-01-28
Migrate from depreciated method for fetching PHPhotos

aarongrider:
Migrate to using `requestImageForAsset`. (https://developer.apple.com/documentation/photokit/phimagemanager/1616964-requestimageforasset?language=objc)

This change fixes the following issues on iOS:
- Importing a photo that has been offloaded to iCloud may fail
- Synchronous import could cause a photo to be imported twice (one low res and one high res)
- Major performance improvements to photo import


ivpusic:
thanks for PR.

I tried to run example project with updates, but image is not rendered anymore after picking single image or cropping it. Could you take a look into this?

aarongrider:
Hey @ivpusic. I think we've decided we have too many changes to contribute back, so I will be just working off a fork and I'll close this out.

Just for the record our app allows import for hundreds of photos at once, and I think we are butting up against some of the limits of image picker. Maybe can contribute back some of these improvements at some point!

Thanks for your consideration.

PR #1541, opened 2021-03-03
Pick current video version instead of original one on iOS

samih7:
### Before
When editing a video with iOS tools (edit, trim) and picking it with `ImagePicker.openPicker`, the original video is retrieved instead of the edited (current) one.

### After 
Given the same scenario, the edited (current) video is retrieved instead of the original one.


jamesone:
Any chances of getting this merged?

Dimasik911:
Any idea when this might be fixed? This is still an issue with the newest version 0.40.2.

Issue #1585, opened 2021-05-13
Cropper won't totally dismiss (frozen UI)


SudoPlz:
### Version
Tell us which versions you are using: 

- react-native-image-crop-picker v0.36.0
- react-native v0.64.0

### Platform
- iOS

### Expected behaviour
openCropper should totally dismiss when hitting Cancel

### Actual behaviour
The openCropper is dismissed, but a grey background remains (UITransitionView) and the actual app view controller (beneath) is frozen

### Steps to reproduce

1. From a modal view controller, `openCamera`, and take a photo

2. then when done, `openCropper`

3. hit cancel

### Attachments

// stacktrace or any other useful debug info

It's a known issue of TOCropViewController https://github.com/TimOliver/TOCropViewController/issues/365

SudoPlz:
Opened a PR to work around the issue here: https://github.com/ivpusic/react-native-image-crop-picker/pull/1586

draperunner:
I think this is a duplicate of https://github.com/ivpusic/react-native-image-crop-picker/issues/1540.

PR #1586, opened 2021-05-13
Making sure the UI doesn't freeze when dismissing the cropper

SudoPlz:
- Fixes: https://github.com/ivpusic/react-native-image-crop-picker/issues/1585


ivpusic:
I tried reproducing issue from https://github.com/ivpusic/react-native-image-crop-picker/issues/1585 with example project, but it works normally. could you retest it to make sure issue is not your-app-specific?

kutsan:
I'm also having exactly the same problem in two of my apps. Pretty sure it's not app specific. Installing the package from this PR solves it for me.

```sh
yarn add 'ivpusic/react-native-image-crop-picker#1586/head'
```

PR #1587, opened 2021-05-13
Now animating when opening the cropper directly

SudoPlz:
Allow the cropper to animate when opened directly (instead of NEVER animating).

We keep the existing behaviour of not animating when opened right after the picker.


ivpusic:
tnx for PR. I tried it, but it currently animates opacity, so it shows different animation than for example opening screen. could you make it the same?

jzxchiang1:
> tnx for PR. I tried it, but it currently animates opacity, so it shows different animation than for example opening screen. could you make it the same?

The animation is controlled in a Pod dependency: https://github.com/TimOliver/TOCropViewController/blob/e0f1d29a4b79f7765b0f4c33a46b099f149c422b/Objective-C/TOCropViewController/TOCropViewController.m#L85

Any way we can patch that from this package?

PR #1595, opened 2021-05-27
fix freeStyleCropEnabled in documentation

christopherklint97:
This PR removes the text `(android only)` in the documentation for freeStyleCropEnabled because this feature actually works on ios.


ivpusic:
thanks

PR #1627, opened 2021-07-20
Fix image compression OOM exception

amiralies:
- Refined target dimension algorithm to ensure both height and width are less than their corresponding max values
- Avoid using separate bitmap instances to allocate less memory
- Use exif orientation instead of bitmap rotation to avoid calling extra `Bitmap.createBitmap`


amiralies:
Looks like we still have issues with very large imgages (for example i've testes with 64MP phone and an instant crash happened)
working on this

amiralies:
Ok, I've just added a commit that uses in sample size when loading a bitmap to load a sampled image and probably consume less memory.

ivpusic:
thanks!

PR #1633, opened 2021-08-05
Android directly open "Gallery"

bluejolly92:
I've edited "PickerModule.java" in order to open directly Gallery instead of default browse


princefishthrower:
@bluejolly92 - what ended up happening here? Would love a `prop` to control which intent is fired - mainly for android devices that have a variety of possible intents. Currently, I think the default photo location that is opened is the downloads folder? Which is often empty for many of our users, or at the very least, a bit confusing...

hoyoungmoon:
Thanks! I use this commit with patch-package and it works well 👍

PR #1643, opened 2021-09-08
add record video in lower quality and record duration limited in second at android.

raindev32:
I already add for feature i founded for make LowQuality in android and maximumVideoDuration for limited record video durations.

=========================
OBJECT PROPS :
=========================
( recordLowQuality?: boolean; )
TRUE for Lower Quality.
FALSE for Default Quality

( maximumVideoDuration?: number; )
Put maximumVideoDuration as props in second value
Example: maximumVideoDuration = 1
=========================

Hope it can help.


rmushtaq123:
@jhontech32 @ivpusic 
can this be merged in the library.
as right now the current andoird recorded video size is of 13+ MB and recorded video is only 10 seconds.

PR #1646, opened 2021-09-16
fix(export): export named functions to match definitions

david50407:
We declared that named exports are exist in [`index.d.ts`](https://github.com/ivpusic/react-native-image-crop-picker/blob/master/index.d.ts#L471-L475), but they are not exist in `index.js`.

This PR exports these named exports in `index.js` that we can really use named import and get benefits with tree-shaking like this: 

```ts
import { openCamera } from 'react-native-image-crop-picker

openCamera()
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,4 +1,12 @@
 import React from 'react';
 
 import {NativeModules} from 'react-native';
-export default NativeModules.ImageCropPicker;
+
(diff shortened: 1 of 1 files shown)


ivpusic (approved):

ivpusic:
thanks

PR #1660, opened 2021-10-06
Add Ukrainian and Swedish languages

echelonka:


lamlengend98:
@ivpusic is Vietnamese availabel??

PR #1671, opened 2021-10-28
[iOS] Keep png format if the input mimeType is image/png

lsdimagine:
This PR is to fix the issue for cropping images with transparent background. Previously the lib always compress image as jpeg so we can't keep the background. This pr returns png image if the specified `mimeType` is `image/png`.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -323,6 +323,8 @@ declare module "react-native-image-crop-picker" {
          * Selected image location
          */
         path: string;
+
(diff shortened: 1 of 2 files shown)


jordangrant:
@mwegener-com No. But this PR works, applied it manually.

mwegener-com:
> @mwegener-com No. But this PR works, applied it manually.

@jordangrant So why is it closed if it's not merged?

PR #1672, opened 2021-11-03
Fc/fix order selection for iOS

fmcruz:
Make the array containing the files the user selected respect the order the user used to add each file


3NM79:
@fmcruz why has this been closed?

PR #1677, opened 2021-11-10
support scenes

brightsider:


bengarney:
This fixes the package being broken on iOS 13; please land it ASAP!

PR #1715, opened 2021-12-31
fix: add max duration video

christianwen:


Naman-Jangid:
@christianwen Is that working ?

Issue #1658, opened 2021-10-04
View blinks on iOS15 if I enable cropping flag


jpudysz:
### Version
- react-native-image-crop-picker v0.36.4
- react-native v0.64.2

### Platform
- iOS

### Expected behaviour
ImagePicker's `ViewController` should fade out gently like on iOS 14 or without `cropping` flag.

### Actual behaviour
View blinks after successful crop which creates poor UX (check videos below).

### Steps to reproduce

1. Open real device/simulator with iOS15

2. Use `openPicker` method with `cropping` flag

3. Select photo and crop it

### Attachments

Example code:
```
ImagePicker
   .openPicker({
        mediaType: 'photo',
        includeBase64: true,
        height: 720,
        width: 1280,
        compressImageQuality: 0.8,
        cropping: true,
        avoidEmptySpaceAroundImage: true
   })
```

iOS 14:

https://user-images.githubusercontent.com/9088288/135848047-991d9d44-7bd2-4777-b504-c4881cdd806e.mov

iOS 15 with `cropping` flag:

https://user-images.githubusercontent.com/9088288/135848060-d2b9f086-99ba-4726-8d38-84a2846b6519.mov

iOS 15 without `cropping` flag:

https://user-images.githubusercontent.com/9088288/135848071-3c4647f0-b2c8-4354-b322-529e702b02f7.mov

tungvm16:
same issue need help

QiYuNew:
I am facing the same issue. RN 0.63, iOS 15.

zumaawad:
Same problem

Milker90:
I found this problem caused by TOCropViewController's property modalPresentationStyle is UIModalTransitionStyleCrossDissolve.

So I fix it like this.

![image](https://user-images.githubusercontent.com/3938006/140458816-f101de7a-3119-4457-a03b-c7b18cb69db9.png)


[patch](https://gist.github.com/Milker90/68b2faf3a7d027422761d43326645c51)

mwegener-com:
It's working great! Thank you

KrifaYounes:
working great.
Submit a pull request

zumaawad:
Is this fix included in the latest build or not yet?

oleksandr-dziuban:
Hi guys, we definitely need a pull request for this and new patch version...
Could someone from maintainer group do that?

zumaawad:
Is this fix included in the latest version or not?

[...]

PR #1739, opened 2022-02-17
fix view blinks on ios15 if enable cropping flag

Milker90:
fix [#1658 ](https://github.com/ivpusic/react-native-image-crop-picker/issues/1658)


ivpusic:
thanks!

PR #1751, opened 2022-03-17
Fix: File provider not found.

maksimko:
Added custom FileProvider implementation to avoid merging collision with another dependencies


ivpusic:
thanks

PR #1770, opened 2022-04-14
fix: image file size for single image

antpuleo2586:
Fixes GIF images reporting incorrect file size.
Have only tested this for single images.


imagina:
The same thing happens when you choose a video file

PR #1779, opened 2022-05-11
ACCESS_MEDIA_LOCATION fix for Android 10+ and uses the new iOS 15 picker with 'any' media type to include both video and photos

bed-sb:
Fixes for:
* Android 10's ACCESS_MEDIA_LOCATION and READ_EXTERNAL_STORAGE permissions
* Fixes the "any" media type so it correctly includes both photos and videos


ddikodroid:
@ivpusic could you review this PR?

PR #1800, opened 2022-06-14
Get correct root VC when using SceneDelegate on iOS 13+

ouabing:
Current version of react-native-image-crop-picker can't present picker when app is using SceneDelegate.

On app using SceneDelegate (for multiple window support or CarPlay support), we should get the root VC from the connected window scene.


ouabing:
Anyone? This is a serious bug for scene based app

ouabing:
ping

PR #1887, opened 2023-01-08
Update README about Android SDK version

pnthach95:
In order to use v0.39.0, user must change SDK version to 33. If not, it throws error at https://github.com/ivpusic/react-native-image-crop-picker/blob/75955f0ef3d3bef13b246523646934cc69f0a604/android/src/main/java/com/reactnative/ivpusic/imagepicker/PickerModule.java#L406

error

### DIFF
--- a/README.md
+++ b/README.md
@@ -270,6 +270,9 @@ android {
 
 Reference for more details https://github.com/ivpusic/react-native-image-crop-picker/issues/1406
 
+- If you use SDK version 33, add following to `app/src/main/AndroidManifest.xml`
+  - `<uses-permission android:name="android.permission.READ_MEDIA_IMAGES"/>`
+
 - [Optional] If you want to use camera picker in your project, add following to `app/src/main/AndroidManifest.xml`
   - `<uses-permission android:name="android.permission.CAMERA"/>`
 
--- a/README.md
+++ b/README.md
@@ -273,7 +273,7 @@ android {
 
 Reference for more details https://github.com/ivpusic/react-native-image-crop-picker/issues/1406
 
-- If you use SDK version 33, add following to `app/src/main/AndroidManifest.xml`
+- If you use SDK version >= 33, add following to `app/src/main/AndroidManifest.xml`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -277,6 +270,9 @@ android {
 
 Reference for more details https://github.com/ivpusic/react-native-image-crop-picker/issues/1406
 
+- If you use SDK version 33, add following to `app/src/main/AndroidManifest.xml`
```
cctv1005s, line 273: Should be SDK version >= 33

README.md
```diff
@@ -277,6 +270,9 @@ android {
...
+  - `<uses-permission android:name="android.permission.READ_MEDIA_IMAGES"/>`
```
cctv1005s, line 274: Do we need `android.permission.READ_EXTERNAL_STORAGE` ?

cctv1005s:
@ivpusic  Could you help to review? This README update is important for users who using Android >= 33

README.md
pnthach95: From [the docs](https://developer.android.com/reference/android/Manifest.permission#READ_EXTERNAL_STORAGE):

> Starting in API level 33, this permission has no effect

So, no, we don't need it

ivpusic (approved):

kwantec:
> In order to use v0.39.0, user must change SDK version to 33. If not, it throws error at
> 
> https://github.com/ivpusic/react-native-image-crop-picker/blob/75955f0ef3d3bef13b246523646934cc69f0a604/android/src/main/java/com/reactnative/ivpusic/imagepicker/PickerModule.java#L406
> 
> error

Hi,  does the component work with newer Android SDK 34 and 35? or must we set it to 33?
I would greatly appreciate it if you can quickly reply to my email daniel@kwantec.com 
thanks :)

pnthach95:
> Hi, does the component work with newer Android SDK 34 and 35? or must we set it to 33?

I'm using with 34. Read README:

> If you use SDK version >= 33

PR #1925, opened 2023-05-23
[Android] Maintain all PNG properties when picking & cropping PNG

GunnarAK:
Hi all,

Please feel free to suggest/make changes as I am not certain this PR is production ready. 

**Let's make this feature request a thing of the past!!**

I'm using the following configuration:
```ts
const response = await ImageCropPicker.openPicker({
  mediaType: 'photo',
  cropping: true,
  enableRotationGesture: false,
  freeStyleCropEnabled: false,
  cropperActiveWidgetColor: theme.colors.accent,
  cropperChooseColor: '#007AFF', // iOS only
  cropperCancelColor: theme.colors.white, // iOS only
  height: 200,
  width: 200,
  compressImageQuality: 0.8,
  showCropFrame: isAndroid,
});
```

With this config I am able to:
- select a PNG image and maintain its transparency
- select a PNG Image, crop it, and maintain its transparency
- select a JPEG image, optionally crop it (just like before)

### DIFF
--- a/android/src/main/java/com/reactnative/ivpusic/imagepicker/Compression.java
+++ b/android/src/main/java/com/reactnative/ivpusic/imagepicker/Compression.java
@@ -64,10 +64,25 @@ class Compression {
             imageDirectory.mkdirs();
         }
 
-        File resizeImageFile = new File(imageDirectory, UUID.randomUUID() + ".jpg");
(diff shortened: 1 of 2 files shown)


amazingSheeppp:
Thank you! That helped me

PR #1974, opened 2023-09-14
Removed camera permission on Android since it is not required

RodolfoGS:
**Manifest.permission.CAMERA** is not required to open the camera and take a picture. So, it could be removed to avoid requesting it to the user.

Also, you need to remove this line from your `app/src/main/AndroidManifest.xml`
```
<uses-permission android:name="android.permission.CAMERA"/>
```

### DIFF
--- a/android/src/main/java/com/reactnative/ivpusic/imagepicker/PickerModule.java
+++ b/android/src/main/java/com/reactnative/ivpusic/imagepicker/PickerModule.java
@@ -309,7 +309,7 @@ class PickerModule extends ReactContextBaseJavaModule implements ActivityEventLi
         setConfiguration(options);
         resultCollector.setup(promise, false);
 
-        permissionsCheck(activity, promise, Arrays.asList(Manifest.permission.CAMERA, Manifest.permission.WRITE_EXTERNAL_STORAGE), new Callable<Void>() {
+        permissionsCheck(activity, promise, Arrays.asList(Manifest.permission.WRITE_EXTERNAL_STORAGE), new Callable<Void>() {
             @Override
             public Void call() {
                 initiateCamera(activity);


1Jesper1:
Interesting, do you have a source where it is explained? Did you test it by reinstalling the app?

RodolfoGS:
> **Note:** If you are using the camera [by invoking an existing camera app](https://developer.android.com/guide/topics/media/camera#camera-apps), your application does not need to request this permission.

Source: https://developer.android.com/guide/topics/media/camera#manifest

Tested on: Android 13 (device), 11 (simulator), 9 (simulator) and 8.1 (device)
All works perfect, uninstalling and installing again to reset the permissions.

1Jesper1:
Thanks for the source. Sounds good!

1Jesper1:
Could someone test this?

RodolfoGS:
I'm using in production without issues so far

1Jesper1:
@RodolfoGS I just sounds to good to be true :) 'Note: If you are using the camera [by invoking an existing camera app](https://developer.android.com/guide/topics/media/camera#camera-apps), your application does not need to request this permission." Do you use another camera app?

RodolfoGS:
Try it yourself, just remove the permission from the code and you will see that the camera app is opened and works perfect.

Also, you could try with **react-native-image-picker** that also do not use the camera permission.

> **Android**
> No permissions required (saveToPhotos requires permission [check](https://github.com/react-native-image-picker/react-native-image-picker#note-on-file-storage)).
> 
> Note: This library does not require Manifest.permission.CAMERA, if your app declares as using this permission in manifest then you have to obtain the permission before using launchCamera.

Source: https://github.com/react-native-image-picker/react-native-image-picker#android-1

PR #2045, opened 2024-04-16
Add Privacy manifest

ujeon:
Add Privacy Manifest for [File timestamp APIs](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api#4278393), .creationDate & .modificationDate

PR #2047, opened 2024-04-17
Adds PrivacyInfo.xcprivacy

Maker-Mark:
## Reference
https://developer.apple.com/documentation/bundleresources/privacy_manifest_files#4284009

## Summary
Adds PrivacyInfor.xcprivacy file with reasons

## Motivation
Apple is requiring all packages have a PrivacyInfor.xcprivacy file that declares the privacy info for each package. Starting May 1st, any app that uses a 3rd party application that is missing declarations will be rejected.


Maker-Mark:
@ivpusic Is this looking right, or are there other ones that should be added?
I'm thinking NSPrivacyAccessedAPICategoryDiskSpace could be another one, but I'm not sure.

Maker-Mark:
https://github.com/ivpusic/react-native-image-crop-picker/pull/2045 is already working on this

PR #2068, opened 2024-06-14
Fixes an issue that would cause the compiler to crash in Xcode 16 beta1

isnine:
### Root cause
Incomplete #import statements cause the compiler to crash when checking for dependencies

image

### Solution

Remove this redundant `#import`.

### DIFF
--- a/ios/src/ImageCropPicker.h
+++ b/ios/src/ImageCropPicker.h
@@ -31,7 +31,6 @@
 #elif __has_include("QBImagePickerController.h") // local QBImagePickerController subspec
 #import "QBImagePickerController.h"
 #else
-#import
 #import "QBImagePicker/QBImagePicker.h"
 #endif
 


isnine:
@ivpusic  Could you help take a look? this is very simple and straightforward fix.

ivpusic:
thanks!

isnine:
@ivpusic Any chance we can tag a 0.41.2?

bite-Horse:
> @isnine Iam Trying to update(pod update) ImagePicker to 0.41.2 using pods but unable to, it always stays on 0.40.0. Even if i add 0.41.2 tag manually, not updating. My pod version is 1.15.2. Since i see the build service crash with Xcode16, trying to update to 0.41.2 to include the fix provided in this PR. Please help me with this.
> 
> ```
> platform :ios, '17.0'
> pod 'RNImageCropPicker', :path => 'react/node_modules/react-native-image-crop-picker'
> ```

After re-installed the react-native-image-crop-picker dependency (verify it in node_modules or package.json), you should cd ios && pod update.

PR #2085, opened 2024-09-17
Do not check legacy permission if targeting Android 13+

maksimko:


pavelbrylov:
@ivpusic could you please take a look at this one, otherwise we have to maintain our own fork, would be great if it's fixed for everyone? Thanks in advance!

Barteque:
Don't you want to modify AndroidManifest.xml so that permission is available for API up to 33 (Adroid 13)?

PR #2110, opened 2024-11-25
Fix openCamera reported dimensions for portrait photos on Android

gaearon:
We're observing width/height being incorrectly reported by the library on Android with portrait photos. This fix, similar to the logic [here](https://github.com/expo/expo/blob/24f5f68e82f527b631c3a19c8a70f06a5ce62968/packages/expo-image-picker/android/src/main/java/expo/modules/imagepicker/exporters/DimensionsExporter.kt#L11-L34), is to swap the reported width/height if the EXIF tag specifies a rotated orientation.

This aligns with the library's behavior on iOS where the width/height are swapped for photos with rotated EXIF.

## Test Plan

Verify that portrait photos (I was using Samsung Galaxy A12) report flipped dimensions, just like the iOS version does.

### DIFF
--- a/android/src/main/java/com/reactnative/ivpusic/imagepicker/PickerModule.java
+++ b/android/src/main/java/com/reactnative/ivpusic/imagepicker/PickerModule.java
@@ -21,6 +21,7 @@ import android.webkit.MimeTypeMap;
 
 import androidx.core.app.ActivityCompat;
 import androidx.core.content.FileProvider;
+import androidx.exifinterface.media.ExifInterface;
 
 import com.facebook.react.bridge.ActivityEventListener;
 import com.facebook.react.bridge.Callback;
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/reactnative/ivpusic/imagepicker/PickerModule.java
```diff
@@ -678,6 +679,15 @@ private WritableMap getImage(final Activity activity, String path) throws Except
             throw new Exception("Cannot select remote files");
         }
         BitmapFactory.Options original = validateImage(path);
+        ExifInterface originalExif = new ExifInterface(path);
```
gaearon, line 682: it's fine to read just the original exif because the orientation tag gets carried forward in the compression step

KasparovNikolay:
Hello, @ivpusic, tell me please, when a new version with this fix will be released?

ivpusic:
released in 0.41.6

Issue #1631, opened 2021-07-29
openCropper is not working for IOS: view is not in the window hierarchy


zvs001:
I see this error in log:
```
[Presentation] Attempt to present
<TOCropViewController: 0x10d126a00> on <PHPickerViewController: 0x10be14b80> (from <PHPickerViewController: 0x10be14b80>) 
whose view is not in the window hierarchy.
```

### Version
- react-native-image-crop-picker v0.36.2
- react-native v0.63.3

### Platform
- iOS (14.5.1)

### Expected behaviour
Modal can be opened.

### Actual behaviour
Modal is not shown. No errors. No callback execution.

### Steps to reproduce

When i call `openCropper()`, it doesn't show modal for some pictures.
Sometimes modal can be opened, but not for same picture.

zvs001:
same issue: https://github.com/ivpusic/react-native-image-crop-picker/issues/1199

Workaround with timeout temporary can help, but it is not stable.

gaearon:
possible fix https://github.com/ivpusic/react-native-image-crop-picker/pull/2118

PR #2118, opened 2024-12-19
Fix image crop modal failing to show

gaearon:
Fixes https://github.com/ivpusic/react-native-image-crop-picker/issues/1631

If the controller is being dismissed, we shouldn't be presenting _from_ that controller.

## Test Plan

No longer getting

```
[UIKitCore] Attempt to present <TOCropViewController: 0x104840200> on
<PHPickerViewController: 0x12bf0ece0> (from <PHPickerViewController: 0x12bf0ece0>) whose
view is not in the window hierarchy.
```

when trying to present the cropper right after the Expo image picker.

See the video in https://github.com/bluesky-social/social-app/pull/7194 to show the interaction. Before the fix, we were getting that log and the cropper would not show up.

### DIFF
--- a/ios/src/ImageCropPicker.m
+++ b/ios/src/ImageCropPicker.m
@@ -126,7 +126,8 @@ RCT_EXPORT_MODULE();
 
 - (UIViewController*) getRootVC {
     UIViewController *root = [[[[UIApplication sharedApplication] delegate] window] rootViewController];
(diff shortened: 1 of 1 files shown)


adrian-koczen:
I tried this fix and it works on iPhone 15 pro & iPhone 8 plus, but for some reason on iPhone 7 not.

PR #2137, opened 2025-02-16
fix: Resolve OutOfMemoryError in image/video processing

GeeksEra:
## Changes
- Implemented memory-efficient image/video processing
- Added proper bitmap recycling
- Optimized large media file handling
- Added memory management safeguards

## Why
The app was experiencing OutOfMemoryError crashes with the following specific error:

This occurred because:
1. Large media files (especially videos) were being loaded into memory without proper size checks
2. Bitmaps weren't being properly recycled after use
3. No memory limit safeguards were in place for large media processing

## Fixes
- Added size checks before loading media files
- Implemented proper bitmap recycling in the processing pipeline
- Added memory-efficient scaling for large media files
- Implemented better error handling for OOM situations

## Testing
- Tested with large video files (>100MB)
- Verified memory usage during multiple media selections
- Confirmed processing of high-resolution images
- Monitored memory allocation during media processing
- Tested on low-memory devices

## Impact
- Reduced memory footprint during media processing
- More stable media handling
- Better user experience with reduced app crashes

### DIFF
--- a/android/src/main/java/com/reactnative/ivpusic/imagepicker/PickerModule.java
+++ b/android/src/main/java/com/reactnative/ivpusic/imagepicker/PickerModule.java
@@ -546,7 +546,8 @@ class PickerModule extends ReactContextBaseJavaModule implements ActivityEventLi
                     public void invoke(Object... args) {
                         String videoPath = (String) args[0];
                         try {
-                            Uri videoUri = Uri.fromFile(new File(videoPath));
+                            File file = new File(videoPath);
+                            Uri videoUri = Uri.fromFile(file);
                             MediaMetadataRetriever retriever = new MediaMetadataRetriever();
                             retriever.setDataSource(activity, videoUri);
                             Bitmap bmp = retriever.getFrameAtTime();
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/reactnative/ivpusic/imagepicker/PickerModule.java
```diff
@@ -563,8 +559,8 @@ public void invoke(Object... args) {
                             video.putInt("size", (int) new File(videoPath).length());
                             video.putInt("duration", (int) duration);
                             video.putString("path", "file://" + videoPath);
-                            video.putString("modificationDate", String.valueOf(modificationDate));
```
ivpusic, line 566: can we get back modificationDate?

GeeksEra: @ivpusic  Done!

ivpusic (approved):

android/src/main/java/com/reactnative/ivpusic/imagepicker/PickerModule.java
MohammedTanveer-Integrass: App getting freeze .Any solution

GeeksEra: @MohammedTanveer-Integrass  I'm not facing this issue, but it might be specific to a android version . Could you share more details?

PR #2147, opened 2025-03-13
Fix the issure in Android where the photo picker does not support a maxFiles selection limit.

xgtz421:
Fix the issure in Android where the photo picker does not support a maxFiles selection limit.
image


ivpusic:
thanks!

PR #2148, opened 2025-03-14
If the "maxFiles" keyword is missing, assign a default value of 5 to align with the documentation.

xgtz421:
Fix the error in the "setConfiguration" function when retrieving "maxFiles" from "options". If the "maxFiles" keyword is missing, assign a default value of 5 to align with the documentation.
image


ivpusic (approved):

ivpusic:
thanks

PR #2164, opened 2025-04-15
Fix maxFiles limit for Android image picker

zahidalidev:
## Description
This PR fixes the `maxFiles` option for Android image picker. Previously, the `maxFiles` option was documented in the README but wasn't properly implemented for Android devices.

## Changes
- Refactored the image picker initialization on Android to use the modern photo picker API
- Added proper implementation of `maxFiles` limit for Android
- Added fallback to the classic picker if the modern picker is not available

## Testing
- Tested on Android 12, 13, and 14
- Verified both single and multiple selection modes work correctly
- Confirmed that the `maxFiles` limit is properly enforced on Android


ivpusic (approved):

ivpusic:
@zahidalidev thanks! could you resolve a merge conflict, and I will merge it afterwards

PR #2170, opened 2025-05-27
feat(android):Target SDK 35 Compatibility (Edge to Edge enforcement) Ucrop Activity

dishantwalia:
After upgrading to target SDK 35, we observed an impact on Android 15 devices, as apps are now edge-to-edge by default if they target Android 15 (API level 35). Image crop picker also impacted with this change as Ucrop activity tool bar distorted and displayed in place of the device status bar and hence user is unable to tap on the icons. Hence removing the style from activity.
Impacted Device : Android 15 & Android 16 with target SDK 35

Before : 
787cefbf-b1e1-42c7-9e77-380f8935a8f3

After: 
Screenshot_20250527_152608

### DIFF
--- a/android/src/main/AndroidManifest.xml
+++ b/android/src/main/AndroidManifest.xml
@@ -23,8 +23,7 @@
         </provider>
 
         <activity
-            android:name="com.yalantis.ucrop.UCropActivity"
-            android:theme="@style/Theme.AppCompat.Light.NoActionBar" />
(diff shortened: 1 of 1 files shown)


HamzaJameel01:
I have updated the library from 0.42.0 to 0.50.1 but still getting same issue in Android 15 @dishantwalia

kontea-dev:
Thanks :)

Issue #2058, opened 2024-05-21
[Android 14] - Limited permissions to gallery not working on Android 14+


Mathitarazonad:
### Version
- react-native-image-crop-picker v0.41.1
- react-native v0.72.13

### Platform
- Android 14

### Expected behaviour
When user grants limited permissions to the gallery, when it only gives access to some photos, when the image picker opens just need to show images that it granted permissions before.

### Actual behaviour
What actually happens is that users provides access to specifics images or videos but when the gallery opens it just shows all images of the device, and not just the specified images that user granted permissions. Also this feature works as expected in iOS but on android it seems to be missing.

### Steps to reproduce

1. In Android 14, request the required permissions to access user media.
![image](https://github.com/ivpusic/react-native-image-crop-picker/assets/111609236/b752cc91-c5fc-474e-8cb6-be7a6a9f246b)

2. User selects the option to select which images are allowed for the app to use.
![image](https://github.com/ivpusic/react-native-image-crop-picker/assets/111609236/4d99c882-3e68-4ec8-aff1-fe511309dd9f)

3. It shows a gallery to allow the user to pick up which images are allowed for the app to use.
![image](https://github.com/ivpusic/react-native-image-crop-picker/assets/111609236/51c83e09-35c7-4b69-9d88-5c7365e31640)

4.  The gallery opens and shows all images not taking into account what user selected previously.
![image](https://github.com/ivpusic/react-native-image-crop-picker/assets/111609236/cf2b9da8-a0e1-4f44-9b25-3cf47f9b9514)
![image](https://github.com/ivpusic/react-native-image-crop-picker/assets/111609236/b7238fa4-fca2-4ef0-a007-d64ad66742d6)

### Attachments
I only can try it with emulators because I don't have physical Android with 14 Api available right now. As I mentioned before, it works well on IOS devices but not on Android 14, also It happens with react-native-image-picker library. Not know exactly what could be missing with Android but I found this documentation in Android developers official page. https://developer.android.com/about/versions/14/changes/partial-photo-video-access?hl=en-419
![image](https://github.com/ivpusic/react-native-image-crop-picker/assets/111609236/a467a228-0787-4602-9e79-93aeee5332f2)

Love react-native-image-crop-picker? Please consider supporting our collective:
👉  https://opencollective.com/react-native-image-crop-picker/donate

deepa98singh:
Same issue. 

Additional behaviour observed,
[...]

PR #2179, opened 2025-07-09
Android gallery partial or limited access 

manacugoze:
- Adding an openAndroidPicker to handle the Android 14 limited/partial access. 
- Fix for https://github.com/ivpusic/react-native-image-crop-picker/issues/2058

https://github.com/user-attachments/assets/c756835f-6599-4c23-937e-3a7d00019b69


harryt04:
@ivpusic can you review this PR? This fixes a crucial bug where Android does not respect the request to only view a limited portion of the device's photos.

manacugoze:
@ivpusic can we review that one before it get's older? We have a ticket blocked with this. Thank you 🙏

PR #2093, opened 2024-09-26
Implement the new official Android Photo Picker

Pauligrinder:
Here's a quick implementation of the new Photo Picker. By using this new picker the READ_MEDIA_IMAGES permission which Google won't allow soon isn't needed anymore.

The implementation may be a tad rough, so maybe someone more seasoned could have a look and improve it if necessary. Mainly I raised some version numbers - they work just fine in the app I'm using it for, but maybe they should be lowered to support more use cases.

PR #2192, opened 2025-08-22
Make "Browse" option available in the Android Photo Picker

marcshilling:
Since adding the official Android photo picker (https://github.com/ivpusic/react-native-image-crop-picker/pull/2093), my users have been complaining that they can't choose images from a cloud provider like Google Drive or Google Photos. This change adds the "Browse" option back to the kebab/3 dot menu on the photo picker:
image

Pulled from this comment: https://stackoverflow.com/a/77317553

### DIFF
--- a/android/src/main/java/com/reactnative/ivpusic/imagepicker/ImageCropPicker.java
+++ b/android/src/main/java/com/reactnative/ivpusic/imagepicker/ImageCropPicker.java
@@ -356,22 +356,23 @@ class ImageCropPicker implements ActivityEventListener {
 
     private void initiatePicker(final Activity activity) {
         try {
-            PickVisualMediaRequest.Builder builder = new PickVisualMediaRequest.Builder();
+            String mimeType;
+
             // Simplified media type handling
             if (mediaType.equals("video")) {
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/reactnative/ivpusic/imagepicker/ImageCropPicker.java
```diff
@@ -356,23 +356,20 @@ private void initiateCamera(Activity activity) {
...
+                intent.setType("image/*");
             } else {
-                builder.setMediaType(ActivityResultContracts.PickVisualMedia.ImageAndVideo.INSTANCE);
+                intent.setType("image/*");
+                intent.putExtra(Intent.EXTRA_MIME_TYPES, new String[]{"image/*", "video/*"});
             }
 
-            Intent intent;
-            if (multiple) {
-                intent = new ActivityResultContracts.PickMultipleVisualMedia(maxFiles).createIntent(activity, builder.build());
```
marcshilling, line 372: Note that we're losing the respect of `maxFiles` here. I don't believe that's possible with this approach.

